PR #1, opened 2019-09-06
Add methods

SergeyTsimfer:
This PR proposes to:

- add convenient properties `crop_shape` and `crop_shape_dice`

- provide methods for showing depth-map, values of cube, correlation, and RGB values along the given horizon, as well as comparing two different horizons with on metrics. `numba` is used where possible

- make `crop` able to use the point not only as the rightmost upper corner, but as any other anchor of the crop. To this end, pass location of the point on the unit cube to `loc` parameter (default behaviour is the same as before)

- better plotters for images in (`iline`, `xline`) coordinates

- add missing docstrings

PR #2, opened 2019-09-06
Side view

SergeyTsimfer:
This PR proposes to:

- change `SeismicGeometry` convertation, so that multiple cubes are created. Each has different orientation: one is `iline`-oriented, `xline`-oriented, `depth`-oriented.

- update `CropBatch` so it is capable of using multiple cubes to accelerate loading of data 

- make `crop` action capable of creating slices of different shapes

- add action to reshape every crop into the same shape

The last two things allow us to load both `iline` and `xline` crops, then force them to have the same shape, so we can use them during training and make model robust to view changes.

This PR is based on #1

### DIFF
--- a/seismiqb/src/crop_batch.py
+++ b/seismiqb/src/crop_batch.py
@@ -629,7 +629,7 @@ class SeismicCropBatch(Batch):
         shape : sequence
             Desired shape of resulting crops.
         """
-        if (np.array(crop.shape) != np.array(shape)).all():
+        if (np.array(crop.shape) != np.array(shape)).any():
             return crop.transpose([1, 0, 2])
         return crop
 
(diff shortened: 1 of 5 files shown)


seismiqb/src/crop_batch.py
```diff
@@ -317,25 +334,63 @@ def _load_cubes_sgy(self, ix, segyfile, dst, src='slices'):
...
+            axis = np.argmin(slice_lens)
+        else:
+            mapping = {0: 0, 1: 1, 2: 2,
+                       'i': 0, 'x': 1, 'h': 2,
+                       'iline': 0, 'xline': 1, 'height': 2, 'depth': 2}
+            axis = mapping[view]
+
+        if axis == 0:
+            crop = self.__load_h5py_i(geom, *slice_)
+        elif axis == 1:
```
akoryagin: Make sure you don't go inside these branches if key `cube_h` or `cube_x` is missing in `geom`

PR #11, opened 2020-02-10
Faster horizons

akoryagin:
**Main additions**
* add function that checks if horizons can be merged together
* add function that merges one horizon into another
* `get_point_cloud` in `SeismicCropBatch` now fetches point clouds from a batch and tries to merge them into `dst`
* add nb measuring the performance of `get_point_cloud`


akoryagin:
[faster horizons with patching](https://trello.com/c/pybJiwqL/4-faster-horizons-with-patching)

seismiqb/src/crop_batch.py
```diff
@@ -327,13 +302,37 @@ def _sgy_init(self, *args, **kwargs):
...
+
+    def _stitch_clouds(self, all_clouds, *args, dst=None, height_margin=2, border_margin=1, **kwargs):
+        """ Stitch a set of point-clouds to a point cloud form dst if possible.
+        Post for `get_point_cloud`-action.
+        """
+        _, _ = args, kwargs
+        # remember, all_clouds contains lists of horizons
+        for horizons_set in all_clouds:
+            for horizon_candidate in horizons_set:
+                flag = False
```
SergeyTsimfer: You can use `for..else` instead of `flag`

seismiqb/src/crop_batch.py
```diff
@@ -509,21 +510,31 @@ def get_point_cloud(self, ix, src_masks='masks', src_slices='slices', dst='predi
...
+        border_margin : int
+            max distance between a pair of horizon-borders when the horizons can be adjacent.
 
         Returns
         -------
         SeismicCropBatch
             batch with fetched labels.
         """
-        _ = dst, to_numba
+        _, _, _ = dst, height_margin, border_margin
```
SergeyTsimfer: You can pass this arguments (as well as theirs respective default values) to the `inbatch_parallel` decorator: this way you will not need to duplicate them in the post function signature

seismiqb/src/crop_batch.py
```diff
@@ -509,21 +510,31 @@ def get_point_cloud(self, ix, src_masks='masks', src_slices='slices', dst='predi
...
 
-        # threshold the mask
-        mask = getattr(self, src_masks)[self.get_pos(None, src_masks, ix)]
+        # threshold the mask, reshape and rotate the mask if needed
+        mask = (getattr(self, src_masks)[self.get_pos(None, src_masks, ix)] > threshold).astype(np.float32)
```
SergeyTsimfer: This type of access to `src_masks` is not how it is intended in `Batch` classes. This is not for your PR, but we will need to fix our components-working everywhere sometime in the future

seismiqb/src/crop_batch.py
```diff
@@ -509,21 +510,31 @@ def get_point_cloud(self, ix, src_masks='masks', src_slices='slices', dst='predi
...
+        mask = np.reshape(mask, mask.shape[:3])
+        mask = np.transpose(mask, axes=order)
 
         # prepare args
+        if isinstance(dst, str):
+            dst = getattr(self, dst)
```
SergeyTsimfer: What if there is no such component/attribute?

seismiqb/src/horizon.py
```diff
@@ -60,6 +60,123 @@ def mask_to_horizon(mask, threshold, averaging, transforms, separate=False):
     return horizons
 
 
+def check_if_joinable(horizon_1, horizon_2, border_margin=1, height_margin=1, diverge_threshold=0.9): #pylint: disable=too-many-branches, too-many-statements
+    """ Check whether a pair of horizons can be stiched together.
+    """
+    # check whether the horizons have overlap in covered area
+    horizon_lesser, horizon_larger = ((horizon_1, horizon_2) if len(horizon_1) <= len(horizon_2)
+                                      else (horizon_2, horizon_1))
+    ctr_common = 0
```
SergeyTsimfer: What is `ctr`?

seismiqb/src/horizon.py
```diff
@@ -60,6 +60,123 @@ def mask_to_horizon(mask, threshold, averaging, transforms, separate=False):
...
+    # no bordering values whatsoever
+    return False
+
+
+def merge_horizon_into_another(horizon_1, horizon_2):
+    """ Merge the first horizon into the second one assuming they can be stitched together.
+        Changes the second horizon inplace.
+    """
+    for key in horizon_1:
+        if key in horizon_2:
```
SergeyTsimfer: Make different `merge_strategies`: one with taking the `mean` of horizons on their overlap, one with keeping the values from original horizon (`horizon_2`)

seismiqb/src/horizon.py
```diff
@@ -60,6 +60,123 @@ def mask_to_horizon(mask, threshold, averaging, transforms, separate=False):
...
+    """ Check whether a pair of horizons can be stiched together.
+    """
+    # check whether the horizons have overlap in covered area
+    horizon_lesser, horizon_larger = ((horizon_1, horizon_2) if len(horizon_1) <= len(horizon_2)
+                                      else (horizon_2, horizon_1))
+    ctr_common = 0
+    for key in horizon_lesser:
+        if horizon_larger.get(key) is not None:
+            ctr_common += 1
+            if not np.isclose(horizon_larger.get(key)[0], horizon_lesser.get(key)[0]):
```
SergeyTsimfer: Add some different strategies of overlap-checking: on maximum difference (that is currently implemented), on average difference, on both of them

seismiqb/src/horizon.py
```diff
@@ -60,6 +60,123 @@ def mask_to_horizon(mask, threshold, averaging, transforms, separate=False):
...
+        for key in ([(iline, min_xline + i) for i in range(-border_margin, 1)]):
+            if key in horizon_larger:
+                ctr_bordering += 1
+                value_min_2 = horizon_larger[key][0]
+                break
+        if np.abs(value_min_1 - value_min_2) <= height_margin:
+            ctr_same += 1
+
+        # check upper adjacency
+        for key in ([(iline, max_xline + i) for i in range(0, border_margin + 1)]):
```
SergeyTsimfer: These two loops defenetely can be merged together somehow

seismiqb/src/horizon.py
```diff
@@ -60,6 +60,123 @@ def mask_to_horizon(mask, threshold, averaging, transforms, separate=False):
...
+        # renew dict values
+        if ilines_to_minmax.get(iline) is None:
+            ilines_to_minmax[iline] = [xline, xline]
+        else:
+            if xline < ilines_to_minmax[iline][0]:
+                ilines_to_minmax[iline][0] = xline
+            elif xline > ilines_to_minmax[iline][1]:
+                ilines_to_minmax[iline][1] = xline
+
+    # form sets of bordering keys for left and right horizon-edge
```
SergeyTsimfer: For each `iline`  that is in horizon, you know `min_xline` and `max_xline`, thus forming complete hull. On the other hand, you know only the leftmost and the rightmost  `xline`s, which can be easily exploited by a complex form of predicted horizon

akoryagin: Not following you. The hull is composed from four sets of keys:
* (iline, max xline) for all ilnes
* (iline, min xline) for all ilines
* (min iline, xline) for all such pairs in the horizon
* (max iline, xline) for all such pairs in the horizon
I cannot think of an easy counterexample - better to discuss this in person for you to draw some pics.

seismiqb/src/horizon.py
akoryagin: true

seismiqb/src/horizon.py
akoryagin: Implemented the same strategy here as in case of non-overlapping horizons

PR #13, opened 2020-04-14
New plotters

akoryagin:
Changes the way how plotting works in `seismiqb`: now every plotting procedure is to be redirected at plotter-classes. There are two of them: `MatplotlibPlotter` and `PlotlyPlotter`. Each plotter-class supports a set of plotting-modes.


seismiqb/src/plotters.py
```diff
@@ -0,0 +1,511 @@
...
+
+
+def filter_kwargs(kwargs, keys):
+    """ Filter the dict of kwargs leaving only supplied keys.
+    """
+    kwargs_ = {}
+    for key in keys:
+        if key in kwargs:
+            kwargs_.update({key: kwargs[key]})
+    return kwargs_
```
SergeyTsimfer, line 48:
```suggestion
   return {key: kwargs[key] for key in keys
           if key in kwargs}
```

seismiqb/src/plotters.py
```diff
@@ -0,0 +1,511 @@
...
+
+def plot_image(image, mode, backend, **kwargs):
+    """ Overall plotter function, redirecting plotting task to one of the methods of backend-classes.
+    """
+    if backend in ('matplotlib', 'plt'):
+        getattr(MatplotlibPlotter(), 'plot_' + mode)(image, **kwargs)
```
SergeyTsimfer, line 54: Both classes already have `Plot..` in their names, so we don't need to add `plot_` prefix to every method

seismiqb/src/plotters.py
```diff
@@ -0,0 +1,511 @@
...
+        # update defaults to make total dict of kwargs
+        defaults = {'xaxis': {'title_text': 'xlines', 'titlefont': {'size': 30}},
+                    'yaxis': {'title_text': 'height', 'titlefont': {'size': 30}, 'autorange': 'reversed'},
+                    'reversescale': True,
+                    'colorscale': 'viridis',
+                    'coloraxis_colorbar': {'title': 'amplitude'},
+                    'opacity' : 1.0,
+                    'title': 'Depth map',
+                    'max_size' : 600}
+        for_update = filter_kwargs(kwargs, list(defaults.keys()) + ['opacity', 'zmin', 'zmax', 'showscale']) # TODO: more args to add in here
```
SergeyTsimfer, line 94: `opacity` is already in `defaults.keys()`; besides, you can actually add everything into the `defaults` dictionary: just set all the parameters to their defaults

seismiqb/src/plotters.py
```diff
@@ -0,0 +1,511 @@
...
+                    'yaxis': {'title_text': 'height', 'titlefont': {'size': 30}, 'autorange': 'reversed'},
+                    'coloraxis_colorbar': {'title': 'depth'},
+                    'title': 'Seismic inline',
+                    'max_size' : 600}
+        grid = (1, len(image))
+        updated = {**defaults, **filter_kwargs(kwargs, list(defaults.keys()))} # TODO: more args to add in here
+
+        # form different groups of kwargs
+        render_kwargs = filter_kwargs(updated, [])
+        label_kwargs = filter_kwargs(updated, ['title'])
```
SergeyTsimfer, line 238: it is the same as `updated['title']`, is not it?

seismiqb/src/plotters.py
```diff
@@ -0,0 +1,511 @@
...
+        width = coeff * width
+        height = coeff * height
+
+        # plot the image and set titles
+        plot_data = go.Image(z=image.T, **render_kwargs)
+        fig = go.Figure(data=plot_data)
+        fig.update_layout(width=width, height=height, **label_kwargs)
+        fig.show()
+
+    def plot_separate(self, image, **kwargs):
```
SergeyTsimfer, line 208: `image` -> `images`?

seismiqb/src/plotters.py
```diff
@@ -0,0 +1,511 @@
...
+            zmax : float
+                the highest brightness-level to be rendered.
+            xaxis : dict
+                controls the properties of xaxis-labels; uses plotly-format.
+            yaxis : dict
+                controls the properties of yaxis-labels; uses plotly-format.
+            other
+        """
+        # update defaults to make total dict of kwargs
+        defaults = {'xaxis': {'title_text': 'xlines', 'titlefont': {'size': 30}},
```
SergeyTsimfer, line 86: We probably should use `Config` for this nested dicts that are ~updateble from the outside

seismiqb/src/plotters.py
```diff
@@ -0,0 +1,511 @@
...
+        # channelize images and put them on a canvas
+        fig, ax = plt.subplots(figsize=updated['figsize'])
+        ax.imshow(image[0].T, **render_kwargs) # note transposition in here
+        ax.set_xlabel(**xaxis_kwargs)
+        ax.set_ylabel(**yaxis_kwargs)
+
+        for img, n_channel in zip(image[1:], (0, 1, 2)):
+            ax.imshow(channelize_image(img.T, total_channels=4, n_channel=n_channel, opacity=updated['opacity']),
+                                       **render_kwargs)
+        fig.suptitle(y=1.1, **label_kwargs)
```
SergeyTsimfer, line 362: Why not use regular `plt.title` here?

seismiqb/src/plotters.py
```diff
@@ -0,0 +1,511 @@
...
+        xaxis_kwargs = filter_kwargs(updated, ['xlabel', 'fontsize', 'family', 'color'])
+        yaxis_kwargs = filter_kwargs(updated, ['ylabel', 'fontsize', 'family', 'color'])
+
+        grid = (1, len(image))
+        fig, ax = plt.subplots(*grid, figsize=updated['figsize'])
+
+        # plot image
+        ax[0].imshow(image[0].T, **render_kwargs)
+        for i in range(1, len(image)):
+            ax[i].imshow(image[i].T, **render_kwargs) # grey colorschemes are embedded here
```
SergeyTsimfer, line 458: Add titles for each of the images

seismiqb/src/plotters.py
```diff
@@ -0,0 +1,511 @@
...
+        for i in range(grid[1]):
+            img = channelize_image(255 * image[i].T, total_channels=4, greyscale=True, opacity=1)
+            fig.add_trace(go.Image(z=img, **render_kwargs), row=1, col=i + 1)
+            fig.update_xaxes(row=1, col=i + 1, **xaxis_kwargs['xaxis'])
+            fig.update_yaxes(row=1, col=i + 1, **yaxis_kwargs['yaxis'])
+        fig.update_layout(**label_kwargs)
+        fig.show()
+
+
+class MatplotlibPlotter:
```
SergeyTsimfer, line 253: We need the ability to save images to a desired location from every method
That (probably) can be done with a fairly simple decorator for class --- one for `plotly`, one for `plt`

seismiqb/src/plotters.py
```diff
@@ -0,0 +1,511 @@
...
+                    'ylabel': 'ilines',
+                    'fontsize': 20,
+                    'fraction': 0.022,
+                    'pad': 0.07,
+                    'labeltop': True,
+                    'labelright': True}
+        updated = {**defaults, **filter_kwargs(kwargs, list(defaults.keys()) + ['family', 'color'])} # TODO: more args to add in here
+
+        # form different groups of kwargs
+        render_kwargs = filter_kwargs(updated, ['cmap'])
```
SergeyTsimfer, line 291: `cmin`, `cmax`?

seismiqb/src/plotters.py
SergeyTsimfer, line 253: In this class, lists of keys that are used for filtration inside functions, are pretty much the same in every method.
`defaults` dict in various methods has a lot of common values as well.
Making them class attributes may have some advantages ranging from reducing duplicate code to ability to create nice and usable docstrings

seismiqb/src/plotters.py
```diff
@@ -0,0 +1,511 @@
...
+    """ Overall plotter function, redirecting plotting task to one of the methods of backend-classes.
+    """
+    if backend in ('matplotlib', 'plt'):
+        getattr(MatplotlibPlotter(), 'plot_' + mode)(image, **kwargs)
+    elif backend in ('plotly', 'go'):
+        getattr(PlotlyPlotter(), 'plot_' + mode)(image, **kwargs)
+    else:
+        raise ValueError('{} backend is not supported!'.format(backend))
+
+class PlotlyPlotter:
```
SergeyTsimfer, line 60: I suggest adding a simple possible version of `transformer` (sic!) from `matplotlib-kwargs` to `plotly-kwargs`: just change `cmin`/`cmax` to `zmin/zmax` and `xlabel/ylabel` to the value `plotly` expects you to use.
You can even apply this conversion at the time of `decorating` methods!

seismiqb/src/plotters.py
SergeyTsimfer: And `cmap` -> `colorscale`

seismiqb/src/plotters.py
SergeyTsimfer: What about `transparent` mode? Can it be done with some set of parameters for other methods?

seismiqb/src/plotters.py
SergeyTsimfer, line 86: Maybe, it would be better to get rid of any seismic-related names, like `ilines/xlines`? They should be passed from the callee here anyways

seismiqb/src/plotters.py
akoryagin: that's right. Reasons it is written in this way: 1) make the code look the same everywhere and 2) add more arguments into the `label`-category if necessary

seismiqb/src/plotters.py
akoryagin: `single` mode in both `plotly` and `matplotlib` backends now supports `opacity`/`alpha` arg, that stands for the transparency of the resulting image

seismiqb/src/plotters.py
akoryagin: why?

seismiqb/src/plotters.py
akoryagin: done

seismiqb/src/plotters.py
akoryagin: done

seismiqb/src/plotters.py
akoryagin: done

seismiqb/src/plotters.py
akoryagin: added two decorators. Needs testing though. Everything works fine with `matplotlib` while `plotly`-saving requires a bunch of other libs.

seismiqb/src/plotters.py
```diff
@@ -64,14 +64,18 @@ def matplotlib_dec(cls):
     # update each method
     for name in names:
         old_ = getattr(cls, name)
```
SergeyTsimfer, line 66: This decorator `wraps` method, so lets call this variable `method`

seismiqb/src/plotters.py
```diff
@@ -64,14 +64,18 @@ def matplotlib_dec(cls):
...
-        def updated(self, old_=old_, **kwargs):
-            # pop savefig-kwarg
-            save = kwargs.pop('save')
-            plt_ = old_(self, **kwargs)
+        def updated(self, *args, old_=old_, **kwargs):
```
SergeyTsimfer, line 67: This decorator wraps method, and we would like to retain docstrings and signature, so lets use `@wraps`

seismiqb/src/plotters.py
```diff
@@ -64,14 +64,18 @@ def matplotlib_dec(cls):
...
+            # pop savefig-kwarg and show-kwarg
+            show = kwargs.pop('show', True)
```
SergeyTsimfer, line 69: Despite our intention to save docstrings and signature, we would like to update them with freshly added arguments of `save` and `show`, so these arguments belong to the signature and don't need weird `pop`s from `kwargs` dictionary

seismiqb/src/plotters.py
```diff
@@ -64,14 +64,18 @@ def matplotlib_dec(cls):
...
+            save = kwargs.pop('save', None)
+            plt_ = old_(self, *args, **kwargs)
```
SergeyTsimfer, line 71: We are getting a `figure`, so lets call this variable `fig` or `figure`

seismiqb/src/plotters.py
```diff
@@ -64,14 +64,18 @@ def matplotlib_dec(cls):
...
             # save if necessary and render
             if save is not None:
                 plt_.savefig(**save)
```
SergeyTsimfer, line 74: I would rather type `plot_image(img, 'overlap', savepath='/a/b/c.png')` than `plot_image(img, 'overlap', save={'fname': '/a/b/c.png', 'bbox_inches': 'tight'})`, so add method (maybe a static one) of saving to the `backend`-class itself, with `kwargs`-parsing and all

seismiqb/src/plotters.py
SergeyTsimfer, line 67: Also, what is the use of `old_=old_`?

seismiqb/src/plotters.py
```diff
@@ -0,0 +1,599 @@
...
+def filter_kwargs(kwargs, keys):
+    """ Filter the dict of kwargs leaving only supplied keys.
+    """
+    kwargs_ = {}
+    for key in keys:
+        if key in kwargs:
+            kwargs_.update({key: kwargs[key]})
+    return kwargs_
+
+def plot_image(image, mode, backend, **kwargs):
```
SergeyTsimfer, line 50: I believe that this function must be using key-word arguments only with the exception of `image` argument, and there should be defaults for them as well

seismiqb/src/crop_batch.py
```diff
@@ -890,14 +890,18 @@ def plot_components(self, *components, idx=0, plot_mode='overlap', order_axes=No
...
-            Opacity for showing images.
         """
-        plot_batch_components(self, *components, idx=idx, plot_mode=plot_mode, order_axes=order_axes, **kwargs)
+        if idx is not None:
+            imgs = [getattr(self, comp)[idx] for comp in components]
+        else:
+            imgs = [getattr(self, comp) for comp in components]
+
+        # set some defaults
+        xlabel = kwargs.pop('xlabel', 'xlines')
```
SergeyTsimfer, line 904:
```suggestion
        kwargs = {
            'xlabel': 'xline',
            ...,
            **kwargs
        }
```

seismiqb/src/cubeset.py
```diff
@@ -556,7 +534,7 @@ def compare_to_labels(self, horizon, src_labels='labels', offset=0, absolute=Tru
                                                                 printer=printer, hist=hist, plot=plot)
 
 
-    def show_slide(self, idx=0, n_line=0, plot_mode='overlap', mode='iline', **kwargs):
+    def show_slide(self, idx=0, n_line=0, mode='iline', plot_mode='overlap', backend='matplotlib', **kwargs):
```
SergeyTsimfer, line 537: rename `mode` to `axis`, `plot_mode` to `mode`

seismiqb/src/cubeset.py
```diff
@@ -568,7 +546,43 @@ def show_slide(self, idx=0, n_line=0, plot_mode='overlap', mode='iline', **kwarg
...
+            Backend to use for render. Can be either 'plotly' or 'matplotlib'. Whenever
+            using 'plotly', also use slices to make the rendering take less time.
         """
         components = ('images', 'masks') if list(self.labels.values())[0] else ('images',)
-        plot_slide(self, *components, idx=idx, n_line=n_line, plot_mode=plot_mode, mode=mode, **kwargs)
+        cube_name = self.indices[idx]
+        geom = self.geometries[cube_name]
+        crop_shape = np.array(geom.cube_shape)
+
+        if mode in ['i', 'il', 'iline']:
```
SergeyTsimfer, line 559: you can parse `mode` (that is to be named `axis`) into integer-axis via `geometry.parse_axis` method, and then use `point[axis+1] = n_line`, `crop_shape[axis] = 1`

seismiqb/src/cubeset.py
```diff
@@ -568,7 +546,43 @@ def show_slide(self, idx=0, n_line=0, plot_mode='overlap', mode='iline', **kwarg
...
+                               .create_masks(dst='masks', width=width, horizons=horizons)
+                               .rotate_axes(src='masks'))
+
+            pipeline = pipeline + labels_pipeline
+
+        batch = (pipeline << self).next_batch(len(self), n_epochs=None)
+        imgs = [np.squeeze(getattr(batch, comp)) for comp in components]
+
+        # configure defaults
+        title = kwargs.pop('title', 'iline {} out of {} on {}'.format(n_line, geom.ilines_len, cube_name))
```
SergeyTsimfer, line 585: update dict

seismiqb/src/cubeset.py
```diff
@@ -568,7 +546,43 @@ def show_slide(self, idx=0, n_line=0, plot_mode='overlap', mode='iline', **kwarg
...
+        order_axes = kwargs.pop('order_axes', (1, 0) if mode in ['i', 'il', 'iline'] else (0, 1))
```
SergeyTsimfer, line 586: use `axis`

seismiqb/src/geometry.py
```diff
@@ -347,22 +348,24 @@ def show_slide(self, loc=None, start=None, end=None, step=1, axis=0, stable=True
         axis = self.parse_axis(axis)
         slide = self.load_slide(loc=loc, start=start, end=end, step=step, axis=axis, stable=stable)
 
-        title = f'{self.index_headers[axis]} {loc} out of {self.lens[axis]}'
-        meta_title = ''
-        plot_images_overlap([slide], title=title, order_axes=order_axes, meta_title=meta_title, **kwargs)
+        # set defaults
+        title = kwargs.pop('title', f'{self.index_headers[axis]} {loc} out of {self.lens[axis]}')
```
SergeyTsimfer, line 352: update dict `kwargs`

seismiqb/src/geometry.py
```diff
@@ -347,22 +348,24 @@ def show_slide(self, loc=None, start=None, end=None, step=1, axis=0, stable=True
...
         data = np.copy(self.trace_container)
         if scaler:
             data = self.scaler(data, mode=scaler)
 
-        title = f'Amplitude distribution for {self.short_name}\nMean/std: {np.mean(data):3.3}/{np.std(data):3.3}'
-        plt.figure()
-        _ = plt.hist(data.ravel(), bins=bins)
-        plt.title(title)
-        plt.show()
+        title = kwargs.pop('title',
```
SergeyTsimfer, line 365: update `kwargs`

seismiqb/src/horizon.py
```diff
@@ -283,11 +282,10 @@ def show_slide(self, loc, width=3, axis=0, stable=True, order_axes=None, **kwarg
         mask = self.add_to_mask(mask, locations, width=width, iterator=iterator if stable else None)
         seismic_slide, mask = np.squeeze(seismic_slide), np.squeeze(mask)
 
-        # Display everything
-        title = f'{self.geometry.index_headers[axis]} {loc} out of {self.geometry.lens[axis]}'
-        meta_title = f'U-horizon {self.name} on {self.geometry.name}'
-        plot_images_overlap([seismic_slide, mask], title=title, order_axes=order_axes, meta_title=meta_title, **kwargs)
-
+        # set defaults if needed and plot the slide
+        title = kwargs.pop('title', (f'U-horizon {self.name} on {self.geometry.name}' + '\n ' +
```
SergeyTsimfer, line 286: update `kwargs`

seismiqb/src/horizon.py
```diff
@@ -1528,34 +1514,54 @@ def show(self, src='matrix', fill_value=None, on_full=True, **kwargs):
         else:
             matrix = copy(matrix).astype(np.float32)
 
+        # defaults for plotting if not supplied in kwargs
+        cmap = kwargs.pop('cmap', 'viridis_r')
```
SergeyTsimfer, line 1518: update `kwargs`

seismiqb/src/horizon.py
```diff
@@ -1528,34 +1514,54 @@ def show(self, src='matrix', fill_value=None, on_full=True, **kwargs):
...
-        plot_image(amplitudes, 'RGB amplitudes of {} on cube {}'.format(self.name, self.cube_name),
-                   rgb=True, **kwargs)
+        amplitudes *= np.asarray(channel_weights).reshape(1, 1, -1)
+
+        # cast values to uint8 if needed
+        if to_uint8:
+            amplitudes = (amplitudes * 255).astype(np.uint8)
+
+        # defaults for plotting if not supplied in kwargs
+        title = kwargs.pop('title', 'RGB amplitudes of {} on cube {}'.format(self.name, self.cube_name))
```
SergeyTsimfer, line 1555: update `kwargs`

seismiqb/src/horizon.py
```diff
@@ -1580,12 +1586,12 @@ def show_slide(self, loc, width=3, axis='i', order_axes=None, heights=None, **kw
             seismic_slide = seismic_slide[:, heights]
             mask = mask[:, heights]
 
-        # Display everything
+        # defaults for plotting if not supplied in kwargs
         header = self.geometry.index_headers[axis]
-        title = f'{header} {loc} out of {self.geometry.lens[axis]}'
-        meta_title = f'Horizon `{self.name}` on `{self.geometry.name}`'
-        plot_images_overlap([seismic_slide, mask], title=title, order_axes=order_axes, meta_title=meta_title, **kwargs)
+        title = kwargs.pop('title', (f'Horizon `{self.name}` on `{self.geometry.name}`' +
```
SergeyTsimfer, line 1591: update `kwargs`

seismiqb/src/plotters.py
```diff
@@ -0,0 +1,661 @@
+""" Plotters-class containing all plotting backend for seismic cubes - data.
```
SergeyTsimfer, line 1: This file has ~nothing to do with seismic data

seismiqb/src/plotters.py
```diff
@@ -0,0 +1,661 @@
...
+    if opacity is not None:
+        background[:, :, -1] = opacity * (image != 0).astype(int)
+
+    return background
+
+
+def filter_kwargs(kwargs, keys):
+    """ Filter the dict of kwargs leaving only supplied keys.
+    """
+    return {key: value for key, value in kwargs.items() if key in keys}
```
SergeyTsimfer, line 44: Why not `key : kwargs[key] for key in keys if key in kwargs`?

seismiqb/src/plotters.py
```diff
@@ -0,0 +1,661 @@
...
+
+def plot_image(image, mode='single', backend='matplotlib', **kwargs):
+    """ Overall plotter function, converting kwarg-names to match chosen backend and redirecting
+    plotting task to one of the methods of backend-classes.
+    """
+    def _convert_kwargs_keys(mode, backend, kwargs):
```
SergeyTsimfer, line 50: Why this function is defined inside other function?

seismiqb/src/plotters.py
```diff
@@ -0,0 +1,661 @@
...
+    """ Plotting backend for plotly.
+    """
+    @staticmethod
+    def save_and_show(fig, **kwargs):
+        """ Save and show plot if needed.
+        """
+        # fetch savefig, show-kwarg and possibly additional kwargs for saving
+        show = kwargs.get('show', True)
+        savepath = kwargs.get('savepath', None)
+        save_kwargs = kwargs.get('save', dict())
```
SergeyTsimfer, line 87: Why are these not arguments of the method?

SergeyTsimfer, line 87: There should be defaults for saving, otherwise created images look horrible

seismiqb/src/plotters.py
SergeyTsimfer: Maybe, it would be better to get rid of any seismic-related names, like ilines/xlines? They should be passed from the callee here anyways

seismiqb/src/plotters.py
```diff
@@ -0,0 +1,661 @@
...
+    """ Plotting backend for matplotlib.
+    """
+    @staticmethod
+    def save_and_show(fig, **kwargs):
+        """ Save and show plot if needed.
+        """
+        # fetch savefig, show-kwarg and possibly additional kwargs for saving
+        show = kwargs.get('show', True)
+        savepath = kwargs.get('savepath', None)
+        save_kwargs = kwargs.get('save', dict())
```
SergeyTsimfer, line 324: Same as `plotly` version

seismiqb/src/plotters.py
```diff
@@ -0,0 +1,661 @@
...
+                yaxis-label.
+            fontsize : int
+                fontsize of labels and titles.
+            curve_labels : list or tuple
+                list/tuple of curve-labels
+            other
+        """
+        # defaults
+        defaults = {'figsize': (8, 5),
+                    'label': 'Loss plot',
```
SergeyTsimfer, line 637: Why are these names here

seismiqb/src/plotters.py
```diff
@@ -0,0 +1,661 @@
...
+        # form groups of kwargs
+        label_kwargs = filter_kwargs(updated, ['label', 'fontsize', 'family', 'color'])
+        xlabel_kwargs = filter_kwargs(updated, ['xlabel', 'fontsize', 'family', 'color'])
+        ylabel_kwargs = filter_kwargs(updated, ['ylabel', 'fontsize', 'family', 'color'])
+
+        # plot the curves
+        plt.figure(figsize=updated['figsize'])
+        curves = plt.plot(*args)
+        if updated['legend']:
+            plt.legend(curves, updated.get('curve_labels', ['Loss ' + str(i) for i in range(len(curves))]))
```
SergeyTsimfer, line 654: This function has nothing to do with loss..

seismiqb/src/plotters.py
```diff
@@ -0,0 +1,661 @@
...
+                    'legend': True}
+        updated = {**defaults, **kwargs}
+
+        # form groups of kwargs
+        label_kwargs = filter_kwargs(updated, ['label', 'fontsize', 'family', 'color'])
+        xlabel_kwargs = filter_kwargs(updated, ['xlabel', 'fontsize', 'family', 'color'])
+        ylabel_kwargs = filter_kwargs(updated, ['ylabel', 'fontsize', 'family', 'color'])
+
+        # plot the curves
+        plt.figure(figsize=updated['figsize'])
```
SergeyTsimfer, line 651: There should be separate dictionaries for `figure` creating and for `legend` creation

seismiqb/src/plotters.py
```diff
@@ -0,0 +1,661 @@
...
+        # form different groups of kwargs
+        render_kwargs = filter_kwargs(updated, [])
+        label_kwargs = filter_kwargs(updated, ['label', 'fontsize', 'family', 'color'])
+        xaxis_kwargs = filter_kwargs(updated, ['xlabel', 'fontsize', 'family', 'color'])
+        yaxis_kwargs = filter_kwargs(updated, ['ylabel', 'fontsize', 'family', 'color'])
+        tick_params = filter_kwargs(updated, ['labeltop', 'labelright'])
+
+        # channelize and plot the image
+        image = channelize_image(image, total_channels=3)
+        plt.figure(figsize=updated['figsize'])
```
SergeyTsimfer, line 495: figure creation should definetely use its own params dictionary, as we want to control its parameters like `resolution`/`face_color` etc

SergeyTsimfer:
Notebook comments:
Why there are so many imports? Lets stop practice of blind copy and pasting from other sources
The order of demonstration should be `geometry -> horizon -> dataset`, and the first two should be created as separate instances
Why there is no `plotly` examples?
Size of most figures is way too big
Formatting of cells is mostly off: there are a lot of lines longer than fits into github
Why the need to compare with old `plot_loss`? Why not compare everything then? Remove it
`show_amplitude_rgb` currently shows pretty much nothing: that is a function to clearly see river estuaries, and I can see none of them
`curve` mode is intended not only for `loss`, but for metrics along lines, so it is a better way of demonstrating it

seismiqb/src/plotters.py
akoryagin: To explicitly show that this function is an important part of `plot_image`-logic. In addition, it's not like we are going to use this function somewhere else

seismiqb/src/cubeset.py
```diff
@@ -556,19 +534,59 @@ def compare_to_labels(self, horizon, src_labels='labels', offset=0, absolute=Tru
...
+        cube_name = self.indices[idx]
+        geom = self.geometries[cube_name]
+        crop_shape = np.array(geom.cube_shape)
+
+        axis = geom.parse_axis(axis)
+        point = np.array([[cube_name, 0, 0, 0]], dtype=object)
+        point[0, axis + 1] = n_line
+        crop_shape[axis] = 1
+
+        pipeline = (Pipeline()
```
SergeyTsimfer, line 564: use `self.p`

seismiqb/src/horizon.py
```diff
@@ -1528,34 +1515,62 @@ def show(self, src='matrix', fill_value=None, on_full=True, **kwargs):
...
+        if to_uint8:
+            amplitudes = (amplitudes * 255).astype(np.uint8)
+
+        # defaults for plotting if not supplied in kwargs
+        kwargs = {
+            'title': 'RGB amplitudes of {} on cube {}'.format(self.name, self.cube_name),
+            'xlabel': 'xlines',
+            'ylabel': 'ilines',
+            **kwargs
+            }
```
SergeyTsimfer, line 1564: Wrong line continuation

seismiqb/src/horizon.py
```diff
@@ -1528,34 +1515,62 @@ def show(self, src='matrix', fill_value=None, on_full=True, **kwargs):
...
+
+        plot_image(amplitudes, mode='rgb', **kwargs)
 
 
     def show_slide(self, loc, width=3, axis='i', order_axes=None, heights=None, **kwargs):
         """ Show slide with horizon on it.
 
+        TODO: add support of order_axes into plotters
```
SergeyTsimfer, line 1572: Remove `TODO`s

seismiqb/src/plotters.py
```diff
@@ -0,0 +1,690 @@
...
+def filter_kwargs(kwargs, keys):
+    """ Filter the dict of kwargs leaving only supplied keys.
+    """
+    return {key: kwargs[key] for key in keys if key in kwargs}
+
+def plot_image(image, mode='single', backend='matplotlib', **kwargs):
+    """ Overall plotter function, converting kwarg-names to match chosen backend and redirecting
+    plotting task to one of the methods of backend-classes.
+    """
+    def _convert_kwargs(mode, backend, kwargs):
```
SergeyTsimfer, line 50: Why this function is inside other function?

PR #27, opened 2020-07-08
Nonrandom support

AlexeyKozhevin:
Nonrandom traces as support

### DIFF
--- a/seismiqb/src/metrics.py
+++ b/seismiqb/src/metrics.py
@@ -141,7 +141,7 @@ class BaseSeismicMetric(Metrics):
         array-like
             Matrix of (n_ilines, n_xlines) shape with computed metric for each point.
         """
-        metric, title = compute_local_corrs(data=self.data, bad_traces=self.bad_traces,
+        metric, title = compute_local_corrs(data=self.data, bad_traces=self.bad_traces, empty_horizon=self.empty_horizon,
                                             kernel_size=kernel_size, reduce_func=reduce_func, **kwargs)
         plot_dict = {
             'spatial': self.spatial,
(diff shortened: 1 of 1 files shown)


seismiqb/src/metrics.py
```diff
@@ -1023,14 +1023,18 @@ def compute_support_func(function_ndarray, function_str, name,
...
                 bad_traces[:, :safe_strip], bad_traces[:, -safe_strip:] = 1, 1
                 bad_traces[:safe_strip, :], bad_traces[-safe_strip:, :] = 1, 1
 
             np.random.seed(0)
             non_zero_traces = np.where(bad_traces == 0)
-            indices = np.random.choice(len(non_zero_traces[0]), supports)
+            if random:
+                indices = np.random.choice(len(non_zero_traces[0]), supports)
+            else:
+                indices = np.arange(0, len(non_zero_traces[0]), len(non_zero_traces[0]) // supports)[:supports]
```
SergeyTsimfer, line 1037: Looks like the indices for different horizons ~mostly will be the same; the problem is, those indices are used to retrieve some of the `non_zero_traces`, which may be completely different for horizons. I think that you need to define a grid beforehand, then use elements from this grid that are present in the labels

seismiqb/src/metrics.py
```diff
@@ -1023,15 +1025,23 @@ def compute_support_func(function_ndarray, function_str, name,
...
                 bad_traces[:, :safe_strip], bad_traces[:, -safe_strip:] = 1, 1
                 bad_traces[:safe_strip, :], bad_traces[-safe_strip:, :] = 1, 1
 
-            np.random.seed(0)
-            non_zero_traces = np.where(bad_traces == 0)
-            indices = np.random.choice(len(non_zero_traces[0]), supports)
-            supports = np.array([non_zero_traces[0][indices], non_zero_traces[1][indices]]).T
+            if random:
+                np.random.seed(0)
+                non_empty_horizon = np.where(empty_horizon == 0)
```
SergeyTsimfer, line 1036: `data` variable contains amplitude values around the horizon and has the shape (# ilines in cube, # crosslines in cube, window). In some points horizon labels are absent: those traces in `data` array are filled with zeroes. Therefore, std of such traces equals 0. That is why the first thing that this function does is computing stds along the last axis of the `data` array and marks traces where it equals to 0 as bad ones

seismiqb/src/metrics.py
roman-kh: Add the capability to generate non-even random traces (e.g. `random` can be a sampler)

PR #28, opened 2020-07-14
Create masks fix

dmylzenova:
Prevent empty masks creation for create_masks call with indices=1.


seismiqb/src/crop_batch.py
```diff
@@ -347,6 +352,8 @@ def create_masks(self, ix, dst, src='slices', width=3, src_labels='labels', indi
 
         for label in labels:
             mask = label.add_to_mask(mask, locations=slice_, width=width)
+            if check_empty and np.sum(mask) > 0.0:
+                return mask
```
SergeyTsimfer, line 356: `break`

seismiqb/src/crop_batch.py
```diff
@@ -331,9 +331,14 @@ def create_masks(self, ix, dst, src='slices', width=3, src_labels='labels', indi
         #pylint: disable=unused-argument
         labels = self.get(ix, src_labels) if isinstance(src_labels, str) else src_labels
         labels = [labels] if not isinstance(labels, (tuple, list)) else labels
+        check_empty = False
 
         if indices in [-1, 'all']:
             indices = np.arange(0, len(labels))
+        elif indices == 1:
```
SergeyTsimfer, line 338: You can do the same in case of any `int`: just check that `sum` of `mask` was changed `int` times

dmylzenova: > You could do exactly the same by using `mask_rebatch`, correct?
> Also, use `git config core.filemode false` in your repository

I was thinking about it but the resulting proportion of non empty masks is too small.  I think it will be too ugly to enlarge initial batch size (and guess batch size after mask rebatch) when we can fix the function.

Also, sometimes one doesn't want to use mask rebatch.

dmylzenova: > You can do the same in case of any `int`: just check that `sum` of `mask` was changed `int` times

ok, I'll do it.

Issue #17, opened 2020-05-14
Further seismiQB improvements 


SergeyTsimfer:
This is the list of things to do in order to make the framework better again:

- Rethink `Horizon` storage location: it would be beneficial to attach them to corresponding `Geometry`, without using additional structure in `Dataset`. It would (hopefully) also make code much cleaner

- Rethink `Sampler` and also make it a property of `Geometry`: also, there is pretty much no need to use [0; 1] sampling anymore

- Add more methods to `UnstructuredHorizon` and make it compatible with 1D index (for 2D seismic)

- Make `Geobody` class work with seismic facies
 
- `metrics.py` currently experiencing a lot of code duplication; also, maybe it would be better to get rid of `evaluate` and use methods directly: that would require additional `plottable` decorator

SergeyTsimfer:
Change `locations` in case of HDF5 file to simple slices:

- split `CropBatch` into separate classes for structured and unstructured geometry, with difference being in how `locations` are working: for the unstructured case, that is a set of trace identifiers, for HDF5 we can use a simple 3D slice

- change `locations` processing in `Horizon` and `SeismicGeometryHDF5`

cdtn:
Allow passing `'all'` to `ilines_range` and `xlines_range` arguments of `SeismicCubeset.make_grid`.

SergeyTsimfer:
add demo notebook for metrics

roman-kh:
> Allow passing `'all'` to `ilines_range` and `xlines_range` arguments of `SeismicCubeset.make_grid`.

remove `_range` for brevity and simplicity

SergeyTsimfer:
Add `GPUCropBatch`

cdtn:
> `SeismicCubeset.load` implements outdated (in many ways) logic and should be updated to the current quality standards.

SergeyTsimfer:
- project: clean-up
- static typing for Horizon and Geometry
- use separate processes for `inference` in `Controllers`

SergeyTsimfer:
- Add `df_transform` to horizon loading

- auto-infer cube step along `INLINE_3D` and  `CROSSLINE_3D` axis

cdtn:
Upgrade `SeismicCubeset.__getitem__` to allow `cubeset[idx]` calls.

AlexeyKozhevin:
Add custom headers creation to `make_sgy`

PR #30, opened 2020-08-18
Better make grid

dmylzenova:
Prevent SeismicCubeset.make_grid from creating crops in places with zero_traces or on any other condition (e.g. create a grid only in places where horizon exists).


seismiqb/src/cubeset.py
```diff
@@ -423,10 +436,18 @@ def _make_axis_grid(axis_range, stride, length, crop_shape):
         grid = []
         for il in ilines:
             for xl in xlines:
+                if np.sum(gaps_matrix[il: il + crop_shape[0],
```
SergeyTsimfer, line 439: Are there situations where we would want to remove crops, that don't consist of only meaningful traces?
I guess if there is an overlapping stride, one can see the possibility to do so, but can you add some other examples of usage?

seismiqb/src/cubeset.py
```diff
@@ -423,10 +436,18 @@ def _make_axis_grid(axis_range, stride, length, crop_shape):
...
+                                      xl: xl + crop_shape[1]]) < gaps_threshold:
+                    continue
                 for h in hs:
                     point = [cube_name, il, xl, h]
                     grid.append(point)
+
         grid = np.array(grid, dtype=object)
+        # Update ranges to actual ones
```
SergeyTsimfer, line 447: I am a user
I ask the `make_grid` method to create a grid for inference for ([100, 500], [600, 700], [800, 1000]) location. I do expect the output array to have the shape of (400, 100, 200).
If you do want to make this optimization, you need to store the left-most upper point and use it later to ~zero-pad your optimized array. Unfortunately, that would not give you any benefits (and even make it almost twice worse memory-wise).

SergeyTsimfer (changes requested):
Overall looks nice, but being 66 commits behind the master branch is not good.
Also, add enhancements to `make_grid`, proposed in #17: better interface for `ilines_range`, `xlines_range`, `heights_range`. And make so we can use `overlap` integer/sequence instead of `strides` tuple.

seismiqb/src/cubeset.py
```diff
@@ -391,9 +392,21 @@ def make_grid(self, cube_name, crop_shape, ilines_range, xlines_range, h_range,
             Distance between grid points.
         batch_size : int
             Amount of returned points per generator call.
+        gaps_matrix : ndarray
```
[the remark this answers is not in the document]
dmylzenova: That is why I didn't call it `filtering_matrix`

seismiqb/src/cubeset.py
dmylzenova: > Are there situations where we would want to remove crops, that don't consist of only meaningful traces?
I was thinking about 2 main use case scenarios:
- Filtering zero traces, we don't want crops that consist of zero traces only. But still, we may want crops from the borders with some part of zero traces. This allowed part can be parametrized with `gaps_threshold` (can add a possibility to make it float as crop_shape proportion in [0, 1], btw)
- Horizon enhancement - I think it is also acceptable to have a part of the crop without horizon because we don't want to skip areas next to horizon borders. It is possible to add some shifts of the crop coordinates if we don't want to allow 
any uncovered places and don't want crops that do not consist of meaningful traces only, but I don't think it is a good idea.

> I guess if there is an overlapping stride, one can see the possibility to do so
?

seismiqb/src/cubeset.py
dmylzenova: I thought it would be a useful optimization since the resulting crop shape is updated in the grid_info.
But I agree it is not quite straightforward, I'll reverse.

PR #31, opened 2020-08-18
Estuaries

akoryagin:
## Additions

* method `Horizon.get_array_values` that allows to get the array of values along the horizon (similar to `Horizon.get_cube_values` but cuts values from an external array).
* Note that the method uses loop for filling in different slices of the resulting array. It is possible to write the function in purely vectorised-manner, but it is 2.0X slower and 2.5X less efficient in terms of memory. One can find the performance test in commit history.


seismiqb/src/horizon.py
```diff
@@ -969,6 +969,38 @@ def get_cube_values(self, window=23, offset=0, scale=False, chunk_size=256):
         background[self.geometry.zero_traces == 1] = np.nan
         return background
 
+    def get_array_values(self, array, array_offset=None, dataset=None, width=5, axes=(2, 1, 0)):
```
SergeyTsimfer, line 972: Seems like you are looking for the same signature, as `Horizon.from_mask` has: `grid_info` or `shifts`.
Also, the word `offset` is pretty much reserved exclusively for the needs of pre-stack geometries.

seismiqb/src/horizon.py
```diff
@@ -969,6 +969,38 @@ def get_cube_values(self, window=23, offset=0, scale=False, chunk_size=256):
...
+        array : np.ndarray
+            A data-array to make a cut from.
+        array_offset : tuple or None
+            an offset defining the location of given array with respect to the horizon.
+            If None, dataset with filled `grid_info` must be supplied.
+        dataset : seismiqb.dataset
+            a dataset with filled grid_info-attr. Must be supplied, if array_offset is None.
+        width : int
+            required width of the resulting cut.
+        axes : tuple
```
SergeyTsimfer, line 986: Do we really need it here? If one wants to transpose, make him do so

seismiqb/src/horizon.py
```diff
@@ -969,6 +969,38 @@ def get_cube_values(self, window=23, offset=0, scale=False, chunk_size=256):
...
+            if not None, axes-transposition with the required axes-order is used.
+        """
+        if array_offset is None:
+            try:
+                grid_info = getattr(dataset, 'grid_info')
+                array_offset = [grid_info['range'][i][0] for i in range(3)]
+            except:
+                raise ValueError('Either array_offset or dataset must be supplied!')
+
+        horizon_offset = self.bbox[0, 0], self.bbox[1, 0]
```
SergeyTsimfer, line 996: `self.i_min`, `self.x_min`

seismiqb/src/utils.py
```diff
@@ -285,6 +285,30 @@ def convert_point_cloud(path, path_save, names=None, order=None, transform=None)
...
+@njit
+def cut_data_along_horizon(array, horizon_matrix, width, horizon_offset, array_offset, fill_value):
+    """ Jit-accelerated function for cutting out array-values along a horizon surface.
+    """
+    # unpack offsets
+    i_hor_offset, x_hor_offset = horizon_offset
+    i_cube_offset, x_cube_offset, h_cube_offset = array_offset
+
+    # loop over elements of cube and fill the cut
+    cut_out = np.zeros(array.shape[:2] + (width, ))
```
SergeyTsimfer, line 297: Default `dtype` is `float64`, which is undesired.
Also, as the word `cutout` is reserved for something else in our line of work, let's use something else.

SergeyTsimfer (changes requested):
I do believe that this `cut_data_along_horizon` function works fine and produces correct results, yet I do have a question.
You have a specification of `WHERE` the array is located (in cubic coordinates): it can be derived from `array_offset` and the shape of a provided array; also, you do know `WHERE` the horizon is located (again, in cubic coordinates). It is exactly the same situation as in `Horizon.add_to_mask` method: we do have two entities that (possible) have an overlap, and we want to add `ones` to the location of a horizon. The only (and I mean it, it is the only) difference is that instead of adding ones to the points where horizon is, you want to retrieve data from an array, so the line 
```
mask[idx_i, idx_x, heights] = alpha
```
in `Horizon.add_to_mask` must be changed to something like
```
background[idx_i, idx_x, np.full_like(heights, j)] = data_chunk[heights, idx_i, idx_x]
```
I copied this line from `Horizon.get_cube_values` method, which does exactly that.

And you can actually easily check if this works: you can create an instance of `SeismicGeometry` from a converted to HDF5 array with predicted estuaries, load horizon, and use `get_cube_values` to obtain the needed result.

Overall, I believe that doing this function with `numpy` approach would be faster and more reasonable at the same time: it also uses the same techniques as pretty much everywhere else in `seismiQB`. Let's leave `numba` as a last resort tool.

seismiqb/src/utils.py
```diff
@@ -285,6 +285,36 @@ def convert_point_cloud(path, path_save, names=None, order=None, transform=None)
...
+    slc_horizon = [slice(l, h) for l, h in zip(heights_start, heights_start + overlap_shape)]
+    overlap_matrix = np.full(array.shape[:2], fill_value=fill_value, dtype=np.float32)
+    overlap_matrix[slc_array] = horizon_matrix[slc_horizon]
+    overlap_matrix -= array_offset[-1]
+
+    # make the cut-array and fill int with array-data located on needed heights
+    result = np.zeros(array.shape[:2] + (width, ))
+    for i, surface_level in enumerate(np.array([overlap_matrix + shift for shift in range(-width // 2 + 1,
+                                                                                          width // 2 + 1)])):
+        mask = (surface_level >= 0) & (surface_level < array.shape[-1]) & (surface_level !=
```
SergeyTsimfer, line 311: You are creating `width` matrices and store them in memory at the same time: that is inefficient 
Then for each of them, you re-compute `mask` and `where` without reusing the data from previous iterations.
Please, move this function to `Horizon` (as it is not `njit`-decorated, you don't need to have it in a separate file), change shape arrays to integers as they are way easier to understand and don't need tons of weird slices and lists, and retrieve heights to get from array once.

seismiqb/src/utils.py
```diff
@@ -285,6 +285,36 @@ def convert_point_cloud(path, path_save, names=None, order=None, transform=None)
...
+
+    # recompute horizon-matrix in array-coordinates
+    slc_array = [slice(l, h) for l, h in zip(overlap_start, overlap_start + overlap_shape)]
+    slc_horizon = [slice(l, h) for l, h in zip(heights_start, heights_start + overlap_shape)]
+    overlap_matrix = np.full(array.shape[:2], fill_value=fill_value, dtype=np.float32)
+    overlap_matrix[slc_array] = horizon_matrix[slc_horizon]
+    overlap_matrix -= array_offset[-1]
+
+    # make the cut-array and fill int with array-data located on needed heights
+    result = np.zeros(array.shape[:2] + (width, ))
```
SergeyTsimfer, line 308: It is possible for value in `array` to be `0`, so let's make this matrix be filled with `nan`s, and dtype `float32`

seismiqb/src/horizon.py
```diff
@@ -986,7 +986,7 @@ def get_array_values(self, array, array_shift=None, dataset=None, width=5, axes=
                 grid_info = getattr(dataset, 'grid_info')
                 array_shift = [grid_info['range'][i][0] for i in range(3)]
             except:
-                raise ValueError('Either array_shift or dataset must be supplied!')
+                raise ValueError('Either array_shift or dataset with filled grid_info must be supplied!') #pylint: disable=raise-missing-from
```
roman-kh, line 989: Do not disable. Use `from` clause to save the context (see [PEP 3134](https://www.python.org/dev/peps/pep-3134/))

models/get_array_funcs.py
```diff
@@ -0,0 +1,88 @@
...
+def make_test_config():
+    # test offsets
+    horizon_offset = (15, 10)
+    array_offset = (15, 17, 118)
+    width = 15
+
+    # test arrays
+    array = np.arange(1500 * 2000 * 35).reshape((1500, 2000, 35))
+    horizon_matrix = np.tile((125, 126, 127, 128, 129, 130), (700, 500))
+    args = (horizon_matrix, horizon_offset, array, array_offset, width, -999)
```
roman-kh, line 15:
```suggestion
    args = (horizon_matrix, horizon_offset, array, array_offset, width, fill_value)
```

seismiqb/src/horizon.py
```diff
@@ -964,6 +964,66 @@ def get_cube_values(self, window=23, offset=0, scale=False, chunk_size=256):
         background[self.geometry.zero_traces == 1] = np.nan
         return background
 
+
+    def get_array_values(self, array, array_shift=None, dataset=None, width=5, axes=(2, 1, 0)):
```
SergeyTsimfer, line 968: Make ~the same signature, as of `Horizon.from_mask` method: no need for dataset passing

seismiqb/src/horizon.py
```diff
@@ -964,6 +964,66 @@ def get_cube_values(self, window=23, offset=0, scale=False, chunk_size=256):
...
+            If None, dataset with filled `grid_info` must be supplied.
+        dataset : seismiqb.dataset
+            a dataset with filled grid_info-attr. Must be supplied, if array_shift is None.
+        width : int
+            required width of the resulting cut.
+        axes : tuple
+            if not None, axes-transposition with the required axes-order is used.
+        """
+        if array_shift is None:
+            try:
```
SergeyTsimfer, line 986: No need for `try/except`

SergeyTsimfer (approved):

PR #32, opened 2020-08-18
Faults

AlexeyKozhevin:
Add `Fault` class and corresponding functional

* fault can be loaded as sticks from csv-like formats or as points from `.npy` and `.hdf5`
* if fault is represented as sticks, we perform interpolation procedure
* there is necessary methods to split fault predictions into separate faults and filter them by size

Also I've added `compute_cdp_transform`, `create_hdf5` and `compute_attribute` into `SeismicGeometry`


seismiqb/src/geometry.py
```diff
@@ -1039,6 +1046,7 @@ class SeismicGeometryHDF5(SeismicGeometry):
     def __init__(self, path, **kwargs):
         self.structured = True
         self.file_hdf5 = None
+        self.iline_to_cdpx, self.xline_to_cdpy = None, None
```
SergeyTsimfer, line 1049: Remove

seismiqb/src/geometry.py
```diff
@@ -1126,6 +1133,38 @@ def load_slide(self, loc, axis='iline', **kwargs):
             slide = self._cached_load(cube, loc)
         return slide
 
+    def compute_cdp_transform(self, sgy_path, second_trace = 20000):
```
SergeyTsimfer, line 1136: Remove

seismiqb/src/geometry.py
```diff
@@ -565,6 +565,11 @@ def make_sgy(self, path_hdf5=None, path_spec=None, postfix='',
             file_name = os.path.basename(path_segy)
             shutil.make_archive(os.path.splitext(path_segy)[0], 'zip', dir_name, file_name)
 
+    def cdp_to_lines(self, points):
```
SergeyTsimfer, line 568: Group that method with `lines_to_cdp`

seismiqb/src/horizon.py
```diff
@@ -645,9 +646,9 @@ def dict_to_points(dictionary):
         return points
 
 
-    @staticmethod
-    def from_mask(mask, grid_info=None, geometry=None, shifts=None,
-                  mode='mean', threshold=0.5, minsize=0, prefix='predict', **kwargs):
+    @classmethod
+    def from_mask(cls, mask, grid_info=None, geometry=None, shifts=None, mode='mean',
```
SergeyTsimfer, line 650: Are changes to this method still relevant?

seismiqb/src/utils.py
```diff
@@ -470,8 +470,6 @@ def groupby_max(array):
     return output[:position]
 
 
-
-
```
SergeyTsimfer, line 474: 😡

seismiqb/src/fault.py
```diff
@@ -0,0 +1,183 @@
...
+class Fault(Horizon):
+    """ !! """
+    FAULT_STICKS = ['INLINE', 'iline', 'xline', 'cdp_x', 'cdp_y', 'height', 'name', 'number']
+    COLUMNS = ['iline', 'xline', 'height', 'name', 'number']
+
+    def file_to_points(self, path):
+        """ Get point cloud array from file values. """
+        #pylint: disable=anomalous-backslash-in-string
+        df = self.read_file(path)
+        df = self.fix_lines(df)
```
SergeyTsimfer, line 32: Make sure that methods appear in code in the same order

seismiqb/src/fault.py
```diff
@@ -0,0 +1,183 @@
...
+        (mask_i_min, mask_i_max), (mask_x_min, mask_x_max), (mask_h_min, mask_h_max) = mask_bbox
+
+        left =  width // 2
+        right = width - left
+
+        points = self.points
+        for i in range(3):
+            positions = np.arange(locations[i].stop)[locations[i]]
+            points = points[np.isin(points[:, i], positions)]
+        def _extend_line(points):
```
SergeyTsimfer, line 87: You can use `cv.dilate` with (1, 3, 1) kernel to do the same (applied to the `mask` after setting fault locations to `1`'s

seismiqb/src/fault.py
```diff
@@ -0,0 +1,183 @@
...
+            _points = []
+            for x in range(-left, right):
+                for y in range(-left, right):
+                    arr = points + np.array([x, y, 0]).reshape(1, 3)
+                    _points.append(arr)
+            return np.concatenate(_points)
+
+        points = _extend_line(points)
+        points = points - np.array([mask_i_min, mask_x_min, mask_h_min]).reshape(1, 3)
+        points = np.maximum(points, 0)
```
SergeyTsimfer, line 97: It is better to create a filtering mask and remove such points altogether

seismiqb/src/fault.py
```diff
@@ -0,0 +1,183 @@
...
+        return df
+
+    @classmethod
+    def split_file(cls, path, dst='faults'):
+        folder = os.path.dirname(path)
+        faults_folder = os.path.join(folder, dst)
+        if faults_folder and not os.path.isdir(faults_folder):
+            os.makedirs(faults_folder)
+        df = pd.read_csv(path, sep='\s+', names=cls.FAULT_STICKS)
+        def _dump(df):
```
SergeyTsimfer, line 175: move this function into the class itself, or somewhere else: let's try not to define functions inside other functions to avoid problems with multiprocessing

seismiqb/src/fault.py
SergeyTsimfer: Looks like you could benefit off of earlier conversion to `np.ndarray`

seismiqb/src/triangulation.py
```diff
@@ -0,0 +1,196 @@
...
+    shape = np.array([np.max(points[:, i]) - np.min(points[:, i]) for i in range(3)])
+    _points = np.zeros((int((shape[0] + 1) * (shape[1] + 1) * (shape[2] + 1)), 3))
+    i = 0
+    for x in prange(np.min(points[:, 0]), np.max(points[:, 0]+1)):
+        for y in range(np.min(points[:, 1]), np.max(points[:, 1]+1)):
+            for z in range(np.min(points[:, 2]), np.max(points[:, 2]+1)):
+                node = np.array([x, y, z])
+                if distance_to_triangle(points, node) < 0.5 * width:
+                    _points[i] = node
+                    i += 1
```
SergeyTsimfer, line 16: You could use `break` to make those loops stop after distancing away from the fault along each direction

SergeyTsimfer (commented):
Add some comments and improve overall style before proceeding to the next step of working with faults. Please, check the timings on interpolation: I am really curious to see which step takes the most time (and I doubt it is distance computation)

seismiqb/src/fault.py
```diff
@@ -0,0 +1,180 @@
...
+    COLUMNS = ['iline', 'xline', 'height', 'name', 'number']
+
+    def from_file(self, path, transform=True, **kwargs):
+        """ Init from path to either CHARISMA or REDUCED_CHARISMA csv-like file
+        or from .npy file with points. """
+        _ = kwargs
+
+        self.path = path
+        self.name = os.path.basename(path)
+        if '.npy' in path:
```
SergeyTsimfer, line 34: Move this to `Horizon` class

seismiqb/src/fault.py
```diff
@@ -0,0 +1,180 @@
...
+            points = np.load(path, allow_pickle=True)
```
SergeyTsimfer, line 35: It is better to set `allow_pickle` to False, as you know that the `dtype` of points is not `object`

seismiqb/src/utils.py
```diff
@@ -621,3 +623,32 @@ def generate_points(edges, divisors, lengths, indices):
                 idx_copy //= divisor
             low[i, j] = edge[idx_copy % length]
     return low
+
+
+@njit
+def filter_faults(labels, threshold=20):
```
SergeyTsimfer, line 629: Move these functions to `faults.py`

seismiqb/src/cubeset.py
```diff
@@ -156,7 +157,21 @@ def create_labels(self, paths=None, filter_zeros=True, dst='labels', labels_clas
             label_list.sort(key=lambda label: label.h_mean)
             if filter_zeros:
                 _ = [getattr(item, 'filter')() for item in label_list]
-            getattr(self, dst)[ix] = label_list
+            getattr(self, dst)[ix] = [item for item in label_list if len(item.points) > 0]
+
+    def dump_labels(self, path, fmt='npy'):
```
SergeyTsimfer, line 162: Looks like the time for `BaseLabel` abstract class has finally come to describe interface of `Horizon`, `Faults`, `Geobody` in one place. What do you think?

seismiqb/src/fault.py
```diff
@@ -0,0 +1,177 @@
+""" Horizon class and metrics. """
```
SergeyTsimfer, line 1: It is not:)

seismiqb/src/fault.py
```diff
@@ -0,0 +1,177 @@
...
+        self.path = path
+        self.name = os.path.basename(path)
+        ext = os.path.splitext(path)[1]
+        if ext == '.npy':
+            points = np.load(path, allow_pickle=True)
+            self.from_points(points, False, **kwargs)
+        else:
+            points = self.csv_to_points(path, **kwargs)
+            self.from_points(points, transform, **kwargs)
+        self._max = self.points.max(axis=0)
```
SergeyTsimfer, line 37: They are already stored in attributes, aren't they?

seismiqb/src/fault.py
```diff
@@ -0,0 +1,177 @@
...
+        mask_bbox = np.array([[locations[0].start, locations[0].stop],
+                              [locations[1].start, locations[1].stop],
+                              [locations[2].start, locations[2].stop]],
+                             dtype=np.int32)
+        points = self.points
+
+        if (self._max < mask_bbox[:, 0]).any() or (self._min >= mask_bbox[:, 1]).any():
+            return mask
+        cond = np.ones(len(points))
+        for i in range(3):
```
SergeyTsimfer, line 101: It is actually faster to filter points on an individual axis one at a time, like:
```
for i in range(3):
    cond = ...
    points = points[cond]
```

Because each successive evaluation deals with way less points

seismiqb/src/fault.py
SergeyTsimfer: This actually should be even needed, as you already filtering `points` prior to it

seismiqb/src/fault.py
```diff
@@ -0,0 +1,177 @@
...
+            cond = np.logical_and(cond, _cond)
+        points = points[cond]
+        points = points - np.array(mask_bbox[:, 0]).reshape(1, 3)
+        points = np.maximum(points, 0)
+        points = np.minimum(points, np.array(mask.shape) - 1)
+        mask[points[:, 0], points[:, 1], points[:, 2]] = 1
+        return mask
+
+    def fix_lines(self, df):
+        """ Fix broken iline and crossline coordinates. """
```
SergeyTsimfer, line 112: Please, add more comments on what is going on in functions related to loading: the process is quite hard to wrap your head around and contains a lot of tricks

seismiqb/src/cubeset.py
AlexeyKozhevin: Good idea

seismiqb/src/fault.py
AlexeyKozhevin: Yeap, I've already removed it

seismiqb/src/geometry.py
AlexeyKozhevin: But why?

seismiqb/src/horizon.py
AlexeyKozhevin: It's not mine, it's from `master`

seismiqb/src/fault.py
AlexeyKozhevin: I want to use pandas `groupby` in `read_sticks`

scripts/convert_to_hdf5.py
```diff
@@ -23,6 +23,8 @@
 
 
 if __name__ == '__main__':
+    import time
```
SergeyTsimfer, line 26: Is it still needed? One can time the execution of a script by using the `time python3 script.py ...`

seismiqb/src/fault.py
```diff
@@ -0,0 +1,309 @@
...
+            os.makedirs(faults_folder)
+        df = pd.read_csv(path, sep=r'\s+', names=cls.FAULT_STICKS)
+        df.groupby('name').apply(cls.fault_to_csv, folder=folder, dst=dst)
+
+    @classmethod
+    def fault_to_csv(cls, df, folder, dst):
+        """ Save separate fault to csv. """
+        df.to_csv(os.path.join(folder, dst, df.name), sep=' ', header=False, index=False)
+
+def split_faults(array, step=None, overlap=1, pbar=False):
```
SergeyTsimfer, line 186: Please, add some comments in code

seismiqb/src/utils.py
```diff
@@ -641,3 +647,57 @@ def generate_points(edges, divisors, lengths, indices):
...
+                max(i - window[0] // 2, 0):min(i + window[0] // 2 + window[0] % 2, array.shape[0]),
+                max(j - window[1] // 2, 0):min(j + window[1] // 2 + window[1] % 2, array.shape[1]),
+                max(k - window[2] // 2, 0):min(k + window[2] // 2 + window[2] % 2, array.shape[2])
+            ]
+            result[i // stride[0], j // stride[1], k // stride[2]] = fn(region.copy())
+    return result
+
+@njit
+def semblance(region):
+    """ Marfurt semblance. """
```
SergeyTsimfer, line 674: Add links to papers

seismiqb/src/utils.py
```diff
@@ -641,3 +647,57 @@ def generate_points(edges, divisors, lengths, indices):
...
+        return ((np.sum(np.sum(region, axis=0), axis=0)**2).sum()) / denum
+    return 0.
+
+@njit
+def semblance_2(region):
+    """ Marfurt semblance v2. """
+    region = region.reshape(-1, region.shape[-1])
+    covariation = region.dot(region.T)
+    s = 0.
+    for i in prange(covariation.shape[0]): # pylint: disable=not-an-iterable
```
SergeyTsimfer, line 686: Official documentation of `numba` says that `prange` works only in `parallel=True` mode

SergeyTsimfer (commented):
I have not looked at functions to extract points from an array and to dump them into hdf5.
I think that adding newlines to split logical blocks of code and comments to especially hard places would greatly enhance this (already amazing) PR

seismiqb/src/cubeset.py
```diff
@@ -358,6 +376,18 @@ def show_slices(self, idx=0, src_sampler='sampler', n=10000, normalize=False, sh
         plot_image(background, **kwargs)
         return batch
 
+    def show_points(self, idx=0, **kwargs):
+        """ Plot 2D map of points. """
+        _map = np.zeros(self.geometries[idx].cube_shape[:-1])
```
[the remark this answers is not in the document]
AlexeyKozhevin: Renamed to `map_`

SergeyTsimfer (commented):
Notebook comments:
You are using quite small batches of size `32`, which are further rebatched. What is the actual `batch_size` seen by the model?
I would also change the width of created masks: it is currently very hard for you model to detect mask, so it makes prediction very thick to boost its accuracy.
I think it might be a good idea to change the `upsampling` strategy of your models: the current one produces a lot of grid-like artifacts.
Loss graphs are looking very suspicious: they stop at really high values of `dice=0.8` at best, which is not good.
Also, I would leave only a few of them, that ready to be published. In the meanwhile, we can merge the rest of the PR into master.

seismiqb/src/geometry.py
```diff
@@ -568,6 +574,133 @@ def make_sgy(self, path_hdf5=None, path_spec=None, postfix='',
...
+            os.remove(path_hdf5)
+
+        with h5py.File(path_hdf5, "a") as file_hdf5:
+            cube_hdf5 = file_hdf5.create_dataset('cube', self.cube_shape)
+            cube_hdf5_x = file_hdf5.create_dataset('cube_x', self.cube_shape[[1, 2, 0]])
+            cube_hdf5_h = file_hdf5.create_dataset('cube_h', self.cube_shape[[2, 0, 1]])
+            _chunks = tqdm(chunks, total=total) if pbar else chunks
+
+            for (iline, xline, height), chunk in _chunks:
+                _slice = (
```
SergeyTsimfer, line 693: `slc`

seismiqb/src/geometry.py
```diff
@@ -568,6 +574,133 @@ def make_sgy(self, path_hdf5=None, path_spec=None, postfix='',
...
+    def create_hdf5(self, path_hdf5, src, shape=None, stride=None, pbar=False):
+        """ Create hdf5 file from np.ndarray or with geological attribute.
+
+        Parameters
+        ----------
+        path_hdf5 : str
+
+        src : np.ndarray or str
+            if `str`, must be a name of the attribute to compute.
+        shape : int, tuple or None
```
SergeyTsimfer, line 647: From the name alone it is quite hard to determine what this argument is doing; rename, maybe?

seismiqb/src/geometry.py
```diff
@@ -568,6 +574,133 @@ def make_sgy(self, path_hdf5=None, path_spec=None, postfix='',
...
+            points = list(itertools.product(*[range(cube.shape[i]) for i in range(3)]))
+        if isinstance(points, list):
+            points = np.array(points)
+
+        attr = attr_filter(cube, result, window, stride, points, attribute)
+        if np.any(stride > 1):
+            attr = zoom(attr, np.array(cube.shape) / np.array(attr.shape))
+        return attr
+
+    def create_hdf5(self, path_hdf5, src, shape=None, stride=None, pbar=False):
```
SergeyTsimfer, line 638: I would rather make this function with `np.ndarray` and `generator` as `src`, and make the generator creation for desired attribute in a separate method

seismiqb/src/geometry.py
```diff
@@ -1162,6 +1301,39 @@ def load_slide(self, loc, axis='iline', **kwargs):
             slide = self._cached_load(cube, loc)
         return slide
 
+    def compute_cdp_transform(self, sgy_path, second_trace=20000):
```
SergeyTsimfer, line 1304: Is that still needed?

seismiqb/src/horizon.py
```diff
@@ -1580,6 +1584,50 @@ def dump(self, path, transform=None, add_height=True):
...
+            shape = (self.i_length, self.x_length, self.h_max - self.h_min + 1)
+            fault_array = np.zeros(shape)
+
+            points = self.points - np.array([self.i_min, self.x_min, self._h_min])
+            fault_array[points[:, 0], points[:, 1], points[:, 2]] = 1
+
+            cube_hdf5[self.i_min:self.i_max+1, self.x_min:self.x_max+1, self.h_min:self.h_max+1] += fault_array
+
+            cube_hdf5_x[
+                self.x_min:self.x_max+1,
```
SergeyTsimfer, line 1616: Create those slices once

seismiqb/src/utils.py
```diff
@@ -641,3 +647,59 @@ def generate_points(edges, divisors, lengths, indices):
...
+def local_correlation(region):
+    """ Correlation in window. """
+    center = region[
+        region.shape[0] - region.shape[0] // 2,
+        region.shape[1] - region.shape[1] // 2,
+    ]
+    corr = np.zeros((region.shape[0], region.shape[1]))
+    for i in prange(region.shape[0]): # pylint: disable=not-an-iterable
+        for j in range(region.shape[1]):
+            corr[i, j] = np.corrcoef(center, region[i, j])[0, 1]
```
SergeyTsimfer, line 704: Compute `corrcoef` by using `numpy` vectorization: that would be way faster

seismiqb/src/triangulation.py
```diff
@@ -0,0 +1,220 @@
+""" Triangulation functions. """
+import numpy as np
+from numba import njit, prange
+
+@njit
```
SergeyTsimfer, line 5: That one is missing `parallel=True`

seismiqb/src/triangulation.py
```diff
@@ -0,0 +1,220 @@
...
+def triangle_rasterization(points, width=1):
+    """ Transform triangle to surface of the fixed thickness.
+
+    Parameters
+    ----------
+    points : numpy.ndarray
+        array of size 3 x 3: each row is a vertex of triangle
```
SergeyTsimfer, line 12: Is it `3 x 3`?

seismiqb/src/triangulation.py
```diff
@@ -0,0 +1,220 @@
...
+    width : int
+        thickness
```
SergeyTsimfer, line 14: Thicc

seismiqb/src/triangulation.py
SergeyTsimfer: Oh. Rename it to `triangle`-related thing

seismiqb/src/triangulation.py
```diff
@@ -0,0 +1,220 @@
...
+
+@njit
+def distance_to_triangle(triangle, node):
+    """ https://gist.github.com/joshuashaffer/99d58e4ccbd37ca5d96e """
+    # pylint: disable=invalid-name, too-many-nested-blocks, too-many-branches, too-many-statements
+    B = triangle[0, :]
+    E0 = triangle[1, :] - B
+    E1 = triangle[2, :] - B
+    D = B - node
+    a = np.dot(E0, E0)
```
SergeyTsimfer, line 66: Out of curiosity: this invokes heavy `np.dot` to make 3 multiplications and 2 summations. Is that actually faster than doing it by hand?

seismiqb/src/cubeset.py
```diff
@@ -156,7 +157,24 @@ def create_labels(self, paths=None, filter_zeros=True, dst='labels', labels_clas
             label_list.sort(key=lambda label: label.h_mean)
             if filter_zeros:
                 _ = [getattr(item, 'filter')() for item in label_list]
-            getattr(self, dst)[ix] = label_list
+            getattr(self, dst)[ix] = [item for item in label_list if len(item.points) > 0]
+
+    def dump_labels(self, path, fmt='npy', separate=False):
```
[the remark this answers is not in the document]
AlexeyKozhevin: Yeap

seismiqb/src/cubeset.py
```diff
@@ -156,7 +157,24 @@ def create_labels(self, paths=None, filter_zeros=True, dst='labels', labels_clas
...
+        """ Dump points to file. """
+        for i in range(len(self.indices)):
+            for label in self.labels[i]:
+                dirname = os.path.dirname(self.index.get_fullpath(self.indices[i]))
+                if path[0] == '/':
```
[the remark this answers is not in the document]
AlexeyKozhevin: I've choose the same way to define path as in `load` and `create_labels`

seismiqb/src/cubeset.py
```diff
@@ -358,6 +376,20 @@ def show_slices(self, idx=0, src_sampler='sampler', n=10000, normalize=False, sh
         plot_image(background, **kwargs)
         return batch
 
+    def show_points(self, idx=0, **kwargs):
+        """ Plot 2D map of points. """
+        map_ = np.zeros(self.geometries[idx].cube_shape[:-1])
```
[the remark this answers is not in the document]
AlexeyKozhevin: Pylint doesn't like `map` name. And me too, I prefer to distinguish my names from reserved words even in short methods

seismiqb/src/fault.py
```diff
@@ -0,0 +1,295 @@
...
+    -------
+    numpy.ndarray
+        array of shape (N, 4) where the first 3 columns are coordinates of points and the last one
+        is for labels
+    """
+    if isinstance(array, SeismicGeometry):
+        array = array.file_hdf5['cube']
+    if chunk_size is None:
+        chunk_size = len(array)
+    chunks = [(start, array[start:start+chunk_size+overlap]) for start in range(0, array.shape[0], chunk_size)]
```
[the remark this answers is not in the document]
AlexeyKozhevin: Agree

seismiqb/src/fault.py
```diff
@@ -0,0 +1,295 @@
...
+    for start, item in chunks:
+        chunk_labels, new_objects = measurements.label(item, structure=s) # compute labels for new chunk
+        chunk_labels[chunk_labels > 0] += n_objects # shift all values to avoid intersecting with previous labels
+        new_overlap = chunk_labels[:overlap]
+
+        if len(prev_overlap) > 0:
+            coords = np.where(prev_overlap > 0)
+            if len(coords[0]) > 0:
+                # while there are the same objects with different labels repeat procedure
+                while (new_overlap != prev_overlap).any():
```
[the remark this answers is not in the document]
AlexeyKozhevin: I didn't check it. But I will

seismiqb/src/fault.py
```diff
@@ -0,0 +1,295 @@
...
+    Returns
+    -------
+    sizes : numpy.ndarray
+    """
+    indices = np.unique(labels[:, 3])
+    sizes = np.zeros_like(indices)
+    for i in prange(len(indices)): # pylint: disable=not-an-iterable
+        label = indices[i]
+        array = labels[labels[:, 3] == label]
+        sizes[label-1] = ((array[:, 0].max() - array[:, 0].min()) ** 2 + (array[:, 1].max() - array[:, 1].min())) ** 0.5
```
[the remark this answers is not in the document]
AlexeyKozhevin: Oops, thanks

seismiqb/src/crop_batch.py
```diff
@@ -295,7 +299,7 @@ def _correct_point_to_grid(self, point, shape, grid_src='quality_grid', eps=3):
 
     @action
     @inbatch_parallel(init='indices', post='_assemble', target='for')
-    def load_cubes(self, ix, dst, src_locations='locations', src_geometry='geometries', **kwargs):
+    def load_cubes(self, ix, dst, src_locations='locations', src_geometry='geometries', use_3d=False, **kwargs):
```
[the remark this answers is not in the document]
AlexeyKozhevin: Renamed to 'slicing'

seismiqb/src/crop_batch.py
```diff
@@ -171,13 +171,17 @@ def crop(self, points, shape=None, direction=(0, 0, 0), side_view=False,
         # pylint: disable=protected-access
 
         # Create all the points and shapes
-        if adaptive_slices:
+        if isinstance(shape, dict):
```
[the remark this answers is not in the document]
AlexeyKozhevin: I added this feature in order to make crops for different cubes along different axes (along faults).
We discussed that I also can do it through changes in sampling.

seismiqb/src/cubeset.py
```diff
@@ -156,7 +157,24 @@ def create_labels(self, paths=None, filter_zeros=True, dst='labels', labels_clas
...
+        """ Dump points to file. """
+        for i in range(len(self.indices)):
+            for label in self.labels[i]:
+                dirname = os.path.dirname(self.index.get_fullpath(self.indices[i]))
+                if path[0] == '/':
+                    path = path[1:]
+                dirname = os.path.join(dirname, path)
+                if not os.path.exists(dirname):
+                    os.makedirs(dirname)
+                if separate:
```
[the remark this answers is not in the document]
AlexeyKozhevin: `separate == True` means that each fault will be saved into a separate fault.

seismiqb/src/fault.py
```diff
@@ -0,0 +1,295 @@
...
+    s = np.ones((3, 3, 3))
+    if pbar:
+        chunks = tqdm(chunks)
+    for start, item in chunks:
+        chunk_labels, new_objects = measurements.label(item, structure=s) # compute labels for new chunk
+        chunk_labels[chunk_labels > 0] += n_objects # shift all values to avoid intersecting with previous labels
+        new_overlap = chunk_labels[:overlap]
+
+        if len(prev_overlap) > 0:
+            coords = np.where(prev_overlap > 0)
```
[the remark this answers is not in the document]
AlexeyKozhevin: The speed is the same. For example, for matrix of shape `(10000, 1000, 100)`:

`np.where`: 2.25 s ± 7.87 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
`np.nonzero`: 2.26 s ± 9.98 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

seismiqb/src/geometry.py
```diff
@@ -568,6 +574,133 @@ def make_sgy(self, path_hdf5=None, path_spec=None, postfix='',
...
             shutil.make_archive(os.path.splitext(path_segy)[0], 'zip', dir_name, file_name)
 
+    def cdp_to_lines(self, points):
+        """ Convert CDP to lines. """
+        inverse_matrix = np.linalg.inv(self.rotation_matrix[:, :2])
+        lines = (inverse_matrix @ points.T - inverse_matrix @ self.rotation_matrix[:, 2].reshape(2, -1)).T
+        return np.rint(lines)
+
+    def compute_attribute(self, locations=None, points=None, window=10, stride=1, attribute='semblance'):
+        """ Compute attribute on cube.
```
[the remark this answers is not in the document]
AlexeyKozhevin: `apply_conv`?

seismiqb/src/triangulation.py
AlexeyKozhevin: Very smeshno

seismiqb/src/geometry.py
```diff
@@ -568,6 +574,133 @@ def make_sgy(self, path_hdf5=None, path_spec=None, postfix='',
...
+
+        cube = self.file_hdf5['cube'][locations[0], locations[1], locations[2]]
+        window = np.minimum(np.array(window), cube.shape)
+
+        shape = np.ceil(np.array(cube.shape) / np.array(stride)).astype(int)
+        result = np.empty(shape)
+        result[:] = np.nan
+
+        if points is None:
+            points = list(itertools.product(*[range(cube.shape[i]) for i in range(3)]))
```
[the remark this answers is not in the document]
AlexeyKozhevin: Can we get the same result shorter than
`np.stack(np.meshgrid(*[range(cube.shape[i]) for i in range(3)]), axis=-1).reshape(-1, 3)`?

seismiqb/src/triangulation.py
AlexeyKozhevin: I removed `prange` because I can't use 'parallel=True': I have `i += 1` inside of loop.

seismiqb/src/triangulation.py
AlexeyKozhevin: But function is calles `**triangle**_rasterization`

seismiqb/src/geometry.py
```diff
@@ -568,6 +574,133 @@ def make_sgy(self, path_hdf5=None, path_spec=None, postfix='',
...
+            if `str`, must be a name of the attribute to compute.
+        shape : int, tuple or None
+            shape of chunks.
+        stride : str
+            stride for chunks
+        pbar : bool
+            progress bar
+        """
+        chunks = []
+        def _infer_tuple(value, default):
```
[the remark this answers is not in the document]
AlexeyKozhevin: But it's nice to define chunk shape as `(length, None, None)` instead of `(length, cube.shape[1], cube.shape[2])`, isn't it?

seismiqb/src/geometry.py
AlexeyKozhevin: Renamed to `chunk_shape`

seismiqb/src/triangulation.py
AlexeyKozhevin: `np.dot(E0, E0)`: 871 ns ± 10.3 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
`np.sum(E0 * E0)`: 3.68 µs ± 30.7 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

seismiqb/src/geometry.py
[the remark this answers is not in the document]
AlexeyKozhevin: But actually now method call `attr_filter` and if we will have other kinds of attributes we can change that behaviour.

seismiqb/src/crop_batch.py
```diff
@@ -304,11 +308,19 @@ def load_cubes(self, ix, dst, src_locations='locations', src_geometry='geometrie
             Component of batch with positions of crops to load.
         dst : str
             Component of batch to put loaded crops in.
+        slicing : str
```
SergeyTsimfer, line 311: I am fine with the current naming scheme; yet, in my head `native` is the one that is implemented in our class (`load_crop`); the second option is `hdf5` that invokes the default `HDF5` data retriever (`getitem` of our geometry)

seismiqb/src/cubeset.py
```diff
@@ -1,10 +1,14 @@
...
+import os
 from glob import glob
+import contextlib
 
 import numpy as np
+import h5py
+from tqdm.auto import tqdm
 
-from ..batchflow import FilesIndex, DatasetIndex, Dataset, Sampler, Pipeline
+from ..batchflow import FilesIndex, DatasetIndex, Dataset, Sampler, Pipeline, D
```
SergeyTsimfer, line 11: You can use `len(self)` instead of `D.size`

seismiqb/src/cubeset.py
```diff
@@ -797,3 +825,83 @@ def assemble_crops(self, crops, grid_info='grid_info', order=None, fill_value=0)
...
+                for i_min, x_min in chunk_grid:
+                    i_max = min(i_min+chunk_shape[0], cube_shape[0])
+                    x_max = min(x_min+chunk_shape[1], cube_shape[1])
+                    self.make_grid(
+                        self.indices[idx], crop_shape,
+                        [i_min, i_max], [x_min, x_max], [0, geometry.depth-1],
+                        strides=crop_stride, batch_size=batch_size
+                    )
+                    chunk_pipeline = pipeline << self
+                    for _ in range(self.grid_iters):
```
SergeyTsimfer, line 897: The loop can be changed to `chunk_pipeline.run(len(self), n_iters=self.grid_iters)`

seismiqb/src/cubeset.py
```diff
@@ -797,3 +825,83 @@ def assemble_crops(self, crops, grid_info='grid_info', order=None, fill_value=0)
...
+        chunk_grid = [
+            make_axis_grid((0, cube_shape[i]), chunk_stride[i], cube_shape[i], crop_shape[i])
+            for i in range(2)
+        ]
+        chunk_grid = np.stack(np.meshgrid(*chunk_grid), axis=-1).reshape(-1, 2)
+
+        if os.path.exists(path_hdf5):
+            os.remove(path_hdf5)
+
+        if pbar:
```
SergeyTsimfer, line 871: Maybe the bar over `chunks` would be better? And it is definitely easier to implement

seismiqb/src/utils.py
```diff
@@ -641,3 +674,57 @@ def generate_points(edges, divisors, lengths, indices):
                 idx_copy //= divisor
             low[i, j] = edge[idx_copy % length]
     return low
+
+# @njit(parallel=True)
```
SergeyTsimfer, line 678: Is that intended?

seismiqb/src/geometry.py
SergeyTsimfer: Remove

SergeyTsimfer (approved):
Fix the remaining comments and, if possible, extend docstrings and code comments in hard places

seismiqb/src/cubeset.py
AlexeyKozhevin: But it is not too detailed

seismiqb/src/cubeset.py
AlexeyKozhevin: I really want to have a progress bar which is one for all chunks and has information about pipeline iterations.
I can add an additional action to pipeline to update the progress bar but imho the current version is more clear.

seismiqb/src/utils.py
AlexeyKozhevin: Nope

seismiqb/src/fault.py
```diff
@@ -0,0 +1,295 @@
...
+            os.makedirs(dst)
+        df = pd.read_csv(path, sep=r'\s+', names=cls.FAULT_STICKS)
+        df.groupby('name').apply(cls.fault_to_csv, folder=folder, dst=dst)
+
+    @classmethod
+    def fault_to_csv(cls, df, dst):
+        """ Save separate fault to csv. """
+        df.to_csv(os.path.join(dst, df.name), sep=' ', header=False, index=False)
+
+def split_faults(array, chunk_size=None, overlap=1, pbar=False):
```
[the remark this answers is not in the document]
AlexeyKozhevin: I've tried but that version is faster

SergeyTsimfer (approved):

PR #34, opened 2020-08-22
add hdf5 to sgy function

dmylzenova:
This PR proposes making inference on a whole cube while dumping results to hdf5 and converting hdf5 to segy.


seismiqb/src/geometry.py
```diff
@@ -787,6 +787,46 @@ def make_hdf5(self, path_hdf5=None, postfix=''):
...
                     file_hdf5['/info/' + attr] = getattr(self, attr)
 
+    # Convert HDF5 to SEG-Y
+    def make_sgy(self, path_hdf5, path_sgy):
+        """ Save `.hdf5` cube without headers in `.segy` format with current geometry headers.
+
+        Parameters
+        ----------
+        path_hdf5 : str
+            Path to load hdf5 file from. File must have a `cube` key where cube data is stored.
```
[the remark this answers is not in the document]
dmylzenova: Why?

seismiqb/src/geometry.py
```diff
@@ -787,6 +787,46 @@ def make_hdf5(self, path_hdf5=None, postfix=''):
...
+
+                with segyio.create(path_save, spec) as dst:
+                    # Copy all textual headers, including possible extended
+                    for i in range(1 + src.ext_headers):
+                        dst.text[i] = src.text[i]
+
+                    c = 0
+                    for i, il in tqdm(enumerate(spec.ilines)):
+                        for x, xl in enumerate(spec.xlines):
+                            idx = self.dataframe['trace_index'][(il, xl)]
```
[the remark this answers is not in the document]
dmylzenova: This method belongs to SeismicGeometrySEGY which loads a dataframe with initialization, doesn't it?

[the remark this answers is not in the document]
SergeyTsimfer: Sorry, I did not notice this as I was sure this method belongs to `SeismicGeometryHDF5`; yet I think that it is a better place for it.
To this end, I've added `segy_path` attribute which is preserved after converting to `HDF5`; this allows you to retrieve the spec of the original file. Also, I believe that you don't need this `dataframe` as you don't need `idx`

SergeyTsimfer (changes requested):

seismiqb/src/geometry.py
```diff
@@ -455,6 +457,60 @@ def show_amplitude_hist(self, scaler=None, bins=50, **kwargs):
         plot_image(data, backend='matplotlib', bins=bins, mode='histogram', **kwargs)
 
 
+    # Convert HDF5 to SEG-Y
+    def make_sgy(self, hdf5_path='temp.hdf5', segy_path=None, dst=None, postfix='', remove_hdf5=False, zip_result=True):
```
SergeyTsimfer, line 461: `path_hdf5`, `path_segy`, `path_spec`?

seismiqb/src/geometry.py
```diff
@@ -455,6 +457,60 @@ def show_amplitude_hist(self, scaler=None, bins=50, **kwargs):
...
+        """ Save `.hdf5` cube without headers in `.segy` format with current geometry headers.
```
SergeyTsimfer, line 462:
```
Convert HDF5 cube to SEG-Y format with current geometry spec
```

seismiqb/src/geometry.py
```diff
@@ -455,6 +457,60 @@ def show_amplitude_hist(self, scaler=None, bins=50, **kwargs):
...
+            Postfix to add to the name of resulting cube.
+        """
+        dst = dst or (os.path.splitext(hdf5_path)[0] + postfix + '.sgy')
+        if not segy_path:
+            if hasattr(self, 'segy_path'):
+                segy_path = self.segy_path
+            else:
+                segy_path = os.path.splitext(self.path) + '.sgy'
+
+        with h5py.File(hdf5_path, 'r') as src:
```
SergeyTsimfer, line 482: Newlines are a blessing

seismiqb/src/geometry.py
```diff
@@ -455,6 +457,59 @@ def show_amplitude_hist(self, scaler=None, bins=50, **kwargs):
...
+        hdf5_path : str
+            Path to load hdf5 file from. File must have a `cube` key where cube data is stored.
+        segy_path : str
+            Path to load segy file from.
+        dst : str
+            Path to store converted cube. By default, new cube is stored right next to original.
+        postfix : str
+            Postfix to add to the name of resulting cube.
+        """
+        dst = dst or (os.path.splitext(hdf5_path)[0] + postfix + '.sgy')
```
SergeyTsimfer, line 475: Maybe `os.path.splitext(self.path)`?

seismiqb/src/geometry.py
dmylzenova: If propose `path_segy` instead of `dst` and `path_spec` as a name for segy cube to load spec from - it is a bit confusing, because the default for  `path_spec`  will be  `self.segy_path`

seismiqb/src/geometry.py
dmylzenova: No, because you would rather want to save segy file next to the hdf5 file you convert it from and not to the folder with the seismic cube itself.

seismiqb/src/geometry.py
dmylzenova: Not sure why you said that but I added a blank line here.

seismiqb/src/geometry.py
```diff
@@ -458,42 +458,45 @@ def show_amplitude_hist(self, scaler=None, bins=50, **kwargs):
 
 
     # Convert HDF5 to SEG-Y
-    def make_sgy(self, hdf5_path='temp.hdf5', segy_path=None, dst=None, postfix='', remove_hdf5=False, zip_result=True):
+    def make_sgy(self, path_hdf5='temp.hdf5', path_spec=None, path_segy=None, postfix='',
```
SergeyTsimfer, line 461: Make `path_segy` one of the last arguments, so you know that it is not needed that often;
The default `temp.hdf5` should be in the same directory, as `self`;

SergeyTsimfer (approved):
How did you test this function?

SergeyTsimfer (approved):

PR #36, opened 2020-09-15
Improvements

SergeyTsimfer:
This PR:

- adds a 3D visualization of horizons

- adds method to compute the area of a seismic cube in square kilometers

- changes `local_corrs` metric to `phases` in controllers


seismiqb/src/crop_batch.py
```diff
@@ -289,7 +289,7 @@ def _correct_point_to_grid(self, point, shape, grid_src='quality_grid', eps=3):
 
     @action
     @inbatch_parallel(init='indices', post='_assemble', target='for')
-    def load_cubes(self, ix, dst, src='locations', **kwargs):
+    def load_cubes(self, ix, dst, src='locations', src_geometry='geometries', **kwargs):
```
cdtn, line 292:
```suggestion
    def load_cubes(self, ix, dst, src_locations='locations', src_data='geometries', **kwargs):
```

I suggest using `src_data` instead of `src_geometry`, since for example in case of estuaries `load_crop()` is called upon an instance of `RectifiedAmplitudes`, which is definitely not a geometry in the usual sense.

And replacing `src` with `src_locations` makes arguments' names more consistent.

seismiqb/src/horizon.py
```diff
@@ -1571,6 +1575,109 @@ def show_amplitudes_rgb(self, width=3, channel_weights=(1, 0.5, 0.25), to_uint8=
...
+        """ Interactive 3D plot. Roughly, does the following:
+            - select `n` points to represent the horizon surface
+            - triangulate those points
+            - remove some of the triangles on conditions
+            - use Plotly to draw the tri-surface
+
+        Parameters
+        ----------
+        n : int
+            Number of points to use to create horizon surface.
```
cdtn, line 1589:
```suggestion
            Number of points to use for horizon surface creation.
```

PR #39, opened 2020-09-24
Add rectified surfaces

cdtn:
Introduce additional functionality to work with rectified surfaces:

- `SeisimcCropBatch.load_cubes`  now takes optional `along_nearest_horizon` argument. It allows croping from data cut along the horizon nearest to sampled point from `src_locations`. 
- `SeisimcCropBatch.create_masks`  now takes optional `along_nearest_horizon` argument. It allows mask creation for the horizon nearest to sampled point from `src_locations` attributes.
- `SeismicCubeset.make_grid` now takes optional `mode` argument. When set to `'2d'` it allows creating crops coverage for rectified data.


seismiqb/src/rectified.py
```diff
@@ -0,0 +1,74 @@
...
+            self.load_data()
+        return self._data
+
+    def load_data(self):
+        """"""
+        self._data = IndexedDict({})
+        for horizon in self.horizons:
+            cutout = horizon.get_cube_values(window=self.window, offset=self.offset, scale=self.scale)
+            cutout[horizon.full_matrix == horizon.FILL_VALUE] = np.nan
+            depth_range = range(horizon.h_min, horizon.h_max)
```
SergeyTsimfer, line 41: This is not a good way of separating horizons. Suppose you have two horizons, following each other, but one is 5 units below the other. They would have quite similar (and overlapping) ranges, so you will get an error during the `load_crop` method.

seismiqb/src/rectified.py
```diff
@@ -0,0 +1,74 @@
...
+            self._data[depth_range] = cutout
+
+    def reload_data(self, window=None, offset=None, scale=None):
+        """"""
+        self.window = window if window is not None else self.window
```
SergeyTsimfer, line 46: `window or self.window`

seismiqb/src/rectified.py
```diff
@@ -0,0 +1,74 @@
...
+        self.offset = offset if offset is not None else self.offset
+        self.scale = scale if scale is not None else self.scale
+        self.load_data()
+
+class RectifiedSurface(Horizon):
```
SergeyTsimfer, line 51: I actually think that you don't need this class at all, neither you need the new method of `add_mask` in the `Horizon` class. All you need to do is to add the option of summing the mask along the `h`-axis in the `create_masks` action of `CropBatch`, as you proposed earlier.

seismiqb/src/cubeset.py
```diff
@@ -359,7 +359,8 @@ def show_slices(self, idx=0, src_sampler='sampler', n=10000, normalize=False, sh
         return batch
 
 
-    def load(self, label_dir=None, filter_zeros=True, dst_labels='labels', p=None, bins=None, **kwargs):
+    def load(self, label_dir=None, filter_zeros=True, dst_labels='labels',
```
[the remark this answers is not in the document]
cdtn: https://github.com/gazprom-neft/seismiqb/issues/17#issuecomment-705478606

seismiqb/src/crop_batch.py
```diff
@@ -289,24 +290,58 @@ def _correct_point_to_grid(self, point, shape, grid_src='quality_grid', eps=3):
...
+        src_geometry : str
+            Dataset attribute to load crops from
+            when `along_nearest_horizon` is False.
+        src_locations : str
+            Dataset attribute to load crops from
+            when `along_nearest_horizon` is True.
+        src_labels : str
+            Dataset attribute to load crops from
+            when `along_nearest_horizon` is True.
+        along_nearest_horizon : bool
```
SergeyTsimfer, line 310: Describe the logic in the first part of the docstring, before parameters

seismiqb/src/crop_batch.py
```diff
@@ -289,24 +290,58 @@ def _correct_point_to_grid(self, point, shape, grid_src='quality_grid', eps=3):
...
-        return geometry.load_crop(location, **kwargs)
+        location = self.get(ix, src_locations)
+        if along_nearest_horizon:
+            nearest_horizon = self.get_nearest_horizon(ix, src_labels, location[2])
+            return nearest_horizon.load_crop_along(location, window, offset)
+        else:
+            geometry = self.get(ix, src_data)
+            return geometry.load_crop(location, **kwargs)
+
+    def get_nearest_horizon(self, ix, src_labels, heights_slice):
```
SergeyTsimfer, line 334: closest horizon?

seismiqb/src/crop_batch.py
```diff
@@ -289,24 +290,58 @@ def _correct_point_to_grid(self, point, shape, grid_src='quality_grid', eps=3):
...
+        """Get horizon with its `h_mean` closest to mean of `heights_slice`."""
+        location_h_mean = (heights_slice.start + heights_slice.stop) // 2
+        nearest_horizon_ind = np.argmin([horizon.h_mean - location_h_mean for horizon in self.get(ix, src_labels)])
```
SergeyTsimfer, line 337: you need to take `absolute` value

seismiqb/src/crop_batch.py
```diff
@@ -340,25 +381,31 @@ def create_masks(self, ix, dst, src='locations', width=3, src_labels='labels', i
         labels = self.get(ix, src_labels) if isinstance(src_labels, str) else src_labels
         labels = [labels] if not isinstance(labels, (tuple, list)) else labels
         check_sum = False
+        rectify = False
+
+        location = self.get(ix, src_locations)
+
+        if along_nearest_horizon:
+            labels = [self.get_nearest_horizon(ix, src_labels, location[2])]
+            rectify = True
```
SergeyTsimfer, line 390: Make `rectify` separate parameter; move `along_closest` to the `indices` logic

seismiqb/src/crop_batch.py
```diff
@@ -340,25 +381,31 @@ def create_masks(self, ix, dst, src='locations', width=3, src_labels='labels', i
...
-            check_sum = True
-        elif isinstance(indices, int):
-            raise ValueError('Inidices should be either -1, 1 or a sequence of ints.')
-        elif isinstance(indices, (tuple, list, np.ndarray)):
-            pass
-        labels = [labels[idx] for idx in indices]
-
-        location = self.get(ix, src)
         shape_ = self.get(ix, 'shapes')
         mask = np.zeros((shape_), dtype='float32')
```
SergeyTsimfer, line 405: Create maks with shape based on `rectify`

seismiqb/src/cubeset.py
```diff
@@ -16,7 +16,6 @@
 from .utils import IndexedDict, round_to_array, gen_crop_coordinates
 
 
-
```
SergeyTsimfer, line 19: 😡

seismiqb/src/cubeset.py
```diff
@@ -158,6 +162,15 @@ def create_labels(self, paths=None, filter_zeros=True, dst='labels', labels_clas
                 _ = [getattr(item, 'filter')() for item in label_list]
             getattr(self, dst)[ix] = label_list
 
+    @property
+    def horizons_matrices(self):
+        """"""
+        if self._horizons_matrices is None:
+            self._horizons_matrices = IndexedDict()
+            for ix in self.indices:
+                self._horizons_matrices[ix] = np.stack([horizon.matrix for horizon in self.labels[ix]])
```
SergeyTsimfer, line 171: Horizons can have `matrices` of different shapes

seismiqb/src/cubeset.py
```diff
@@ -423,6 +450,12 @@ def make_grid(self, cube_name, crop_shape, ilines=None, xlines=None, heights=Non
             in a crop in the grid. Default value is 0.
             If float, proportion from the total number of traces in a crop will be computed.
         """
+        if mode == '2d':
+            horizon_num = 0 if horizon_num is None else horizon_num
```
SergeyTsimfer, line 454: Make the default `0`

seismiqb/src/horizon.py
```diff
@@ -858,7 +859,7 @@ def create_sampler(self, bins=None, quality_grid=None, **kwargs):
         self.sampler = HorizonSampler(np.histogramdd(points/self.cube_shape, bins=bins))
 
 
-    def add_to_mask(self, mask, locations=None, width=3, alpha=1, **kwargs):
+    def add_to_mask(self, mask, locations=None, width=3, alpha=1, rectify=False, **kwargs):
```
SergeyTsimfer, line 862: Looks messy

seismiqb/src/horizon.py
```diff
@@ -962,6 +969,27 @@ def get_cube_values(self, window=23, offset=0, scale=False, chunk_size=256):
         background[self.geometry.zero_traces == 1] = np.nan
         return background
 
+
+    """Cached version of `Horizon.get_cube_values`
+    to allow direct calls of that method that won't be cached."""
+    cached_get_cube_values = lru_cache(1)(get_cube_values)
+
+
+    def load_crop_along(self, location, window, offset, **kwargs):
```
SergeyTsimfer, line 978: Add the un-cached version of this method

seismiqb/src/horizon.py
```diff
@@ -962,6 +969,27 @@ def get_cube_values(self, window=23, offset=0, scale=False, chunk_size=256):
...
+        """Make crops from data cut along the horizon.
+
+        Parameters
+        ----------
+        location : iterable of at least 2 slices
```
SergeyTsimfer, line 983: `sequence`; you don't need it to be iterable

SergeyTsimfer (changes requested):

seismiqb/src/crop_batch.py
```diff
@@ -339,26 +386,29 @@ def create_masks(self, ix, dst, src='locations', width=3, src_labels='labels', i
         """
         labels = self.get(ix, src_labels) if isinstance(src_labels, str) else src_labels
         labels = [labels] if not isinstance(labels, (tuple, list)) else labels
+        indices = [indices] if isinstance(indices, int)
```
SergeyTsimfer, line 389: That would break something

seismiqb/src/crop_batch.py
```diff
@@ -321,12 +360,20 @@ def create_masks(self, ix, dst, src='locations', width=3, src_labels='labels', i
...
-            If 1 or 'single', one random label will be added.
-            If array-like then elements are interpreted as indices of the desired labels
-            and must be ints in range [0, len(horizons) - 1].
-            Note if you want to pass an index of a single label it must be a list with one
-            element.
+            If 'all', all labels will be added.
+            If 'single', one random label will be added.
+            If int or array-like then element(s) are interpreted as indices of
+            desired labels and must be ints in range [0, len(horizons) - 1].
+        along_nearest_horizon : bool
```
SergeyTsimfer, line 367: Still think it should just be a mode of `indices` : `nearest`

seismiqb/src/crop_batch.py
```diff
@@ -339,26 +386,29 @@ def create_masks(self, ix, dst, src='locations', width=3, src_labels='labels', i
...
+                labels = [labels[idx] for idx in indices]
+            elif indices == 'single':
+                np.random.shuffle(labels)
+                check_sum = True
+            elif indices == 'all':
+                pass
+
+        rectify = along_nearest_horizon if rectify is None
+        crop_shape = self.get(ix, 'shapes')
+        crop_shape = (crop_shape[0], crop_shape[1], 1) if rectify else crop_shape
```
SergeyTsimfer, line 407: `(*crop_shape[:2], 1)`

seismiqb/src/crop_batch.py
```diff
@@ -515,9 +565,9 @@ def concat_components(self, ix, src, dst, axis=-1):
 
     @action
     @inbatch_parallel(init='indices', target='for', post='_masks_to_horizons_post')
-    def masks_to_horizons(self, ix, src='masks', src_locations='locations', dst='predicted_labels', prefix='predict',
-                          threshold=0.5, mode='mean', minsize=0, mean_threshold=2.0, adjacency=1,
-                          order=(2, 0, 1), skip_merge=False):
+    def masks_to_horizons(self, ix, src_masks='masks', src_locations='locations', dst='predicted_labels',
```
SergeyTsimfer, line 568: Previous line splits actually had meaning behind

seismiqb/src/cubeset.py
```diff
@@ -4,6 +4,8 @@
 
 import numpy as np
 
+from matplotlib import pyplot as plt
```
SergeyTsimfer, line 7: image

seismiqb/src/cubeset.py
```diff
@@ -506,6 +535,43 @@ def _make_axis_grid(axis_range, stride, length, crop_shape):
         }
 
 
+    def show_grid(self, plot_over=0, plot_dict=None):
```
SergeyTsimfer, line 538: This looks exactly like that thing that should be added to the `plotters.py`, not here

seismiqb/src/horizon.py
SergeyTsimfer, line 862: `rectify` - > `2d`
Better yet, `mode = '3d'`

seismiqb/src/horizon.py
```diff
@@ -890,20 +895,23 @@ def add_to_mask(self, mask, locations=None, width=3, alpha=1, **kwargs):
             overlap = self.matrix[i_min - self.i_min:i_max - self.i_min,
                                   x_min - self.x_min:x_max - self.x_min]
```
SergeyTsimfer, line 897: You are doing a lot of unnecessary things in case of `2d` mask; you could return the value right here

seismiqb/src/horizon.py
```diff
@@ -962,6 +970,33 @@ def get_cube_values(self, window=23, offset=0, scale=False, chunk_size=256):
         background[self.geometry.zero_traces == 1] = np.nan
         return background
 
+
+    # Cached version of `Horizon.get_cube_values`.
+    # to allow direct calls of that method that won't be cached.
```
SergeyTsimfer, line 975: Thinking emoji

seismiqb/src/horizon.py
```diff
@@ -962,6 +970,33 @@ def get_cube_values(self, window=23, offset=0, scale=False, chunk_size=256):
...
+        cache_call : bool
+            Whether cache call of `Horizon.get_cube_values` or not.
+            Defaults to False.
+
+        Other parameters are the same as in `Horizon.get_cube_values`.
+        """
+        if cache_call:
+            data = self.cached_get_cube_values(window, offset, **kwargs)
+        else:
+            data = self.get_cube_values(window, offset, **kwargs)
```
SergeyTsimfer, line 996: When I was talking about the non-cached version, I meant using the cube to get the crop, not loading everything along the horizon

seismiqb/src/crop_batch.py
cdtn: Both `load_cubes` and `create_masks` should have same argument indicating that rectified crop creation mode is on.
Naming this arg `indices` is quite confusing.

seismiqb/src/horizon.py
```diff
@@ -886,28 +890,31 @@ def add_to_mask(self, mask, locations=None, width=3, alpha=1, **kwargs):
...
-            heights -= (mask_h_min + low)
-
-            for _ in range(width):
-                mask[idx_i, idx_x, heights] = alpha
-                heights += 1
+        if i_max > i_min and x_max > x_min:
+            overlap = self.matrix[i_min - self.i_min : i_max - self.i_min,
+                                  x_min - self.x_min : x_max - self.x_min]
+            if mode == '2d':
+                binarize = np.vectorize(lambda x: 0 if x < 0 else alpha)
```
SergeyTsimfer, line 897: You are compiling this function at each call to `add_to_mask`. I believe something like 
```
(overlap >= 0).astype(np.int32) * alpha
```
should do the trick

seismiqb/src/horizon.py
```diff
@@ -959,9 +969,54 @@ def get_cube_values(self, window=23, offset=0, scale=False, chunk_size=256):
...
+            matrix = (matrix - self.h_min) / (self.h_max - self.h_min)
+        matrix[matrix < 0] = 0
+        return matrix
+
+
+    # Cached version of `get_heights_matrix`.
+    cached_get_heights_matrix = lru_cache(1)(get_heights_matrix)
+
+
+    def crop_matrix(self, matrix, location, **kwargs):
```
SergeyTsimfer, line 994: It is hard to guess here or in `CropBatch` actions that `matrix` is actually a string. `src`, maybe?

seismiqb/src/horizon.py
```diff
@@ -959,9 +969,54 @@ def get_cube_values(self, window=23, offset=0, scale=False, chunk_size=256):
...
         return background
 
+
+    # Cached version of `get_cube_values`.
+    cached_get_cube_values = lru_cache(1)(get_cube_values)
+
+
+    def get_heights_matrix(self, scale=True):
+        """Transform `full_matrix` attribute for further cropping."""
+        matrix = self.full_matrix[..., np.newaxis]
```
SergeyTsimfer, line 983: This addition of a new axis should be part of the `load_crops` action

seismiqb/src/crop_batch.py
```diff
@@ -289,24 +290,73 @@ def _correct_point_to_grid(self, point, shape, grid_src='quality_grid', eps=3):
...
+    def get_nearest_horizon(self, ix, src_labels, heights_slice):
+        """Get horizon with its `h_mean` closest to mean of `heights_slice`."""
+        location_h_mean = (heights_slice.start + heights_slice.stop) // 2
+        nearest_horizon_ind = np.argmin([abs(horizon.h_mean - location_h_mean) for horizon in self.get(ix, src_labels)])
+        return self.get(ix, src_labels)[nearest_horizon_ind]
+
+
+    @action
+    @inbatch_parallel(init='indices', post='_assemble', target='for')
+    def load_crops(self, ix, dst, matrix, src_locations='locations',
```
SergeyTsimfer, line 345: `load_attribute`, maybe?

seismiqb/src/horizon.py
```diff
@@ -377,6 +377,7 @@ def __init__(self, storage, geometry, name=None, **kwargs):
         self._matrix = None
         self._points = None
         self._depths = None
+        self._cube_values_along = None
```
SergeyTsimfer, line 380: Not used anymore

seismiqb/src/horizon.py
```diff
@@ -959,9 +969,54 @@ def get_cube_values(self, window=23, offset=0, scale=False, chunk_size=256):
...
+        if nan_zero_traces:
+            background[self.geometry.zero_traces == 1] = np.nan
         return background
 
+
+    # Cached version of `get_cube_values`.
+    cached_get_cube_values = lru_cache(1)(get_cube_values)
+
+
+    def get_heights_matrix(self, scale=True):
```
SergeyTsimfer, line 981: Make it a property

seismiqb/src/horizon.py
SergeyTsimfer, line 994: Rename to `get_attribute` and default values for arguments: 

- 'amplitudes' for `src`
- `None` for `location`: by default, you produce a map of the desired attribute on the whole cube

seismiqb/src/horizon.py
```diff
@@ -1170,6 +1223,17 @@ def evaluate(self, compute_metric=True, supports=50, plot=True, savepath=None, p
...
 
+    @lru_cache(1)
+    def cached_metrics_evaluate(self, scale=True, metric='support_corrs', supports=50, agg='nanmean', **kwargs):
+        """Cached metrics calcucaltion."""
+        metrics = self.horizon_metrics.evaluate(metric=metric, supports=supports, agg=agg,
+                                                plot=False, savepath=None, **kwargs)
+        if scale:
+            metrics = (metrics + 1) / 2
+            metrics = np.nan_to_num(metrics)
+        return metrics[..., np.newaxis]
```
SergeyTsimfer, line 1234: This new axis should also be in `load_crops` action of `CropBatch`

seismiqb/src/horizon.py
cdtn: IMO, default param for function with `get_attribute` name is kinda counterintuitive. What exactly does one should expect to get when called without any params? I say `ValueError`.

seismiqb/src/horizon.py
cdtn: What for? There is `Horizon.full_matrix` already. The function here fulfills two purposes:
1. Makes heights normalization optional.
2. Makes data access cacheable.

When converted to property, it won't longer be able to be parameterized.
However, I can add the following, if you wish:

`heights_matrix = property(cached_get_heights_matrix)`

seismiqb/src/crop_batch.py
```diff
@@ -98,13 +100,13 @@ def __getattr__(self, name):
         return super().__getattr__(name)
 
     def get(self, item=None, component=None):
-        """ Custom access for batch attribures.
-        If `component` looks like `label` or `geometry` then we retrieve
-        that dictionary from attached dataset and use unsalted
-        version of `item` as key. Otherwise, we get position of `item`
-        in the current batch and use it to index sequence-like `component`.
+        """ Custom access for batch attributes.
+        If `component` has an entry from `DATASET_ATTRIBUTES` than retrieve it
```
SergeyTsimfer, line 104: then

seismiqb/src/crop_batch.py
```diff
@@ -919,12 +978,13 @@ def analytic_transform(self, crop, axis=1, mode='phase'):
             If 'freq', compute instantaneous frequency.
         """
         analytic = hilbert(crop, axis=axis)
-        phase = np.unwrap(np.angle(analytic))
```
SergeyTsimfer, line 922: Current implementation computes the phase or frequency by the well-known formulae. This method has nothing to do with that and should be implemented in a separate method

seismiqb/src/cubeset.py
```diff
@@ -151,12 +157,37 @@ def create_labels(self, paths=None, filter_zeros=True, dst='labels', labels_clas
...
-
             label_list = [labels_class(path, self.geometries[ix], **kwargs) for path in paths[ix]]
             label_list.sort(key=lambda label: label.h_mean)
             if filter_zeros:
                 _ = [getattr(item, 'filter')() for item in label_list]
             getattr(self, dst)[ix] = label_list
+            self.cached_attributes.append(dst)
+
+
+    def reset_caches(self, attrs=None, cubes=None):
```
SergeyTsimfer, line 168: Do you see a use case for the `cubes` argument? Especially knowing that the cache is common for all the instances of, say, `SeismicGeometry`.
Let's not bring entities unless we need them and see their usage

seismiqb/src/cubeset.py
```diff
@@ -423,6 +471,12 @@ def make_grid(self, cube_name, crop_shape, ilines=None, xlines=None, heights=Non
             in a crop in the grid. Default value is 0.
             If float, proportion from the total number of traces in a crop will be computed.
         """
+        if mode == '2d':
+            horizon = self.labels[cube_name][horizon] if isinstance(horizon, int) else horizon
```
SergeyTsimfer, line 475: I think that it is easy enough for a person to pass `horizon.h_mean` to the arguments, and there is no need for `horizon` argument

seismiqb/src/horizon.py
```diff
@@ -886,28 +889,30 @@ def add_to_mask(self, mask, locations=None, width=3, alpha=1, **kwargs):
...
+            overlap = self.matrix[i_min - self.i_min : i_max - self.i_min,
+                                  x_min - self.x_min : x_max - self.x_min]
+            if mode == '2d':
+                mask[i_min - mask_i_min : i_max - mask_i_min,
+                     x_min - mask_x_min : x_max - mask_x_min, 0] = (overlap >= 0).astype(np.int32) * alpha
+            elif mode == '3d':
+                # Coordinates of points to use in overlap local system
+                zero_trace_check = (overlap != self.FILL_VALUE)
+                heights_check = (overlap >= mask_h_min + low) & (overlap <= mask_h_max - high)
+                idx_i, idx_x = np.asarray(zero_trace_check & heights_check).nonzero()
```
SergeyTsimfer, line 902: The previous syntax is looking way better: you can see the checks one after the other

seismiqb/src/horizon.py
```diff
@@ -886,28 +889,30 @@ def add_to_mask(self, mask, locations=None, width=3, alpha=1, **kwargs):
...
+                heights = overlap[idx_i, idx_x]
+
+                # Convert coordinates to mask local system
+                idx_i += i_min - mask_i_min
+                idx_x += x_min - mask_x_min
+                heights -= (mask_h_min + low)
+
+                for offset in range(width):
```
SergeyTsimfer, line 910: `offset` is reserved in seismic applications for something else; use `shift`

seismiqb/src/horizon.py
```diff
@@ -886,28 +889,30 @@ def add_to_mask(self, mask, locations=None, width=3, alpha=1, **kwargs):
...
+                    mask[idx_i, idx_x, heights + offset] = alpha
         return mask
 
 
-    def get_cube_values(self, window=23, offset=0, scale=False, chunk_size=256):
+    def get_cube_values(self, window=23, offset=0, scale=False, chunk_size=256, nan_zero_traces=True, cached=False):
```
SergeyTsimfer, line 915: `cached` is not used?

seismiqb/src/horizon.py
SergeyTsimfer: Are you actually using both versions of this matrix? Scaled and unscaled one
If not, keep one of them, and then it definitely should be a property
And you don't need the `heights` prefix in the name; `scaled_matrix` is fine

seismiqb/src/horizon.py
```diff
@@ -959,9 +967,61 @@ def get_cube_values(self, window=23, offset=0, scale=False, chunk_size=256):
...
+        matrix[matrix < 0] = 0
+        return matrix
+
+
+    # Cached version of `get_heights_matrix`.
+    cached_get_heights_matrix = lru_cache(CACHE_SIZE)(get_heights_matrix)
+
+
+    def get_attribute(self, src_attribute, location=(None, None, None), **kwargs):
+        """Make crops from data cut along the horizon.
```
SergeyTsimfer, line 993: Only knowing in advance what this method is for allows understanding of what is going on in that docstring

seismiqb/src/horizon.py
```diff
@@ -1039,11 +1099,25 @@ def get_cube_values_line(self, orientation='ilines', line=1, window=23, offset=0
 
     # Horizon properties
     @property
-    def amplitudes(self):
+    def cube_values(self):
```
SergeyTsimfer, line 1102: Keep the old name and make a new alias
`cube_values` is very vague and misleading, imo

seismiqb/src/horizon.py
```diff
@@ -1039,11 +1099,25 @@ def get_cube_values_line(self, orientation='ilines', line=1, window=23, offset=0
...
+
+    @property
+    def amplitudes(self):
+        """ Alias for cube values. Depending on what loaded to cube geometries
+        might actually not be amplitudes, so use it with caution.
+        """
+        return self.cube_values
+
+    @property
+    def phases(self):
```
SergeyTsimfer, line 1116: That is as misleading as it gets. While I can see the logic behind adding `cube_values` property, the `geometry` attribute describes, well, geometry of the object: it has nothing to do with the actual values inside the `SEG-Y` or `HDF5` file and the burden of knowing which cube is which is on the user.
As this attribute in 99% of cases does not evaluate to actual `phases_along_horizon`, this should not be here

seismiqb/src/horizon.py
```diff
@@ -1631,6 +1725,13 @@ def show_slide(self, loc, width=3, axis='i', order_axes=None, zoom_slice=None, *
         plot_image([seismic_slide, mask], order_axes=order_axes, **kwargs)
 
 
+    def reset_cache(self):
+        """ Clear cached data. """
+        for attr in dir(self):
+            if attr.startswith('cached'):
```
SergeyTsimfer, line 1731: A better way to check it is to verify that `isinstance(attr, LRUCache)` and then use `reset`

SergeyTsimfer (changes requested):

seismiqb/src/crop_batch.py
```diff
@@ -28,6 +28,8 @@ class SeismicCropBatch(Batch):
         'target': 'for',
         'post': '_assemble'
     }
+    DATASET_ATTRIBUTES = ['label', 'geom', 'fan', 'channel']
```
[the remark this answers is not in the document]
cdtn: I agree that `SeismicCropBatch.get` logic should be modified, but it has nothing to do with current PR, actually

seismiqb/src/crop_batch.py
```diff
@@ -289,44 +291,97 @@ def _correct_point_to_grid(self, point, shape, grid_src='quality_grid', eps=3):
...
+            crop = crop[..., np.newaxis]
+        elif final_ndim != crop.ndim:
+            raise ValueError("Crop returned by `Horizon.get_attribute` has {} dimensions, but shape conversion "
+                             "to expected {} dimensions is not implemented.".format(crop.ndim, final_ndim))
+        return crop
+
+    @action
+    @inbatch_parallel(init='indices', post='_assemble', target='for')
+    def create_masks(self, ix, dst, src_labels='labels', locations='locations',
+                     use_labels='all', mode='auto', thicken=3):
```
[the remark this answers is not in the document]
cdtn: I'm ok with keeping `width` (though naming is kinda vague), but as for `index` — to an external user reading pipeline actions it's totally unobvious what it stands for. Cube index? Label index? Locations index? Or is it about some horizon internal attribute? Definitely should be changed to something more obvious and self-documenting.

seismiqb/src/crop_batch.py
```diff
@@ -289,44 +291,97 @@ def _correct_point_to_grid(self, point, shape, grid_src='quality_grid', eps=3):
...
+        src_attribute : str
+            A keyword defining label attribute to make crops from:
+            - 'cube_values' — cube values cut along the horizon;
+            - 'heights' — matrix of heights with zero-filled nans;
+            - 'metrics' — random support metrics matrix.
+        src_labels : str
+            Dataset attribute with labels dict.
+        locations : str
+            Component of batch that stores locations of crops.
+        final_ndim : 2 or 3
```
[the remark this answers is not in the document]
cdtn: Not self-explaining at all.

seismiqb/src/cubeset.py
cdtn: The fact that cache is shared between class instances is a flaw, not a feature. Therefore it is expected to be fixed sometime. When it does, `reset_caches` should iterate all cubes anyway. I say one rather obvious extra statement does not make the code more difficult to maintain, and at the same time makes it more flexible.

seismiqb/src/horizon.py
cdtn: Having an attribute in `Horizon` that stores values of loaded cube under `amplitudes` name is even more misleading, than having `phases` standing for the same as `amplitudes`. I can only suggest to get rid of both of them, keeping `cube_values` attribute only.

seismiqb/src/horizon.py
cdtn: Current `lru_cache` is inapplicable to properties. So this change will slowdown pipeline run a bit.

seismiqb/src/horizon.py
cdtn: `attr` itself is never instance of `lru_cache`, neither are any of its attributes, added by decoration.
Can do this check instead: `if callable(attr) and hasattr(attr, 'cache')`.

cdtn: Or maybe I shouldn't. Calling `getattr(self, attr)` will invoke every lazy calculation possible.

seismiqb/src/horizon.py
SergeyTsimfer: Why is not it applicable?
Anyways, if you use the `cache` decorator, and only then the `property` one, it will work

cdtn: It doesn't work ¯\\\_(ツ)_/¯. If you first decorate function with `@property` and then with `@lru_cache(1)`, nothing happens. And if `lru_cache(1)` goes before `@property`, than it fails with error due to its implementation details.

As for `cache`, it's introduced in `python 3.9`. And even though it's just a special case of `functools.lru_cache`
```
def cache(user_function, /):
    'Simple lightweight unbounded cache.  Sometimes called "memoize".'
    return lru_cache(maxsize=None)(user_function)
```
substituting our thread-safe `@lru_cache` with `@functools.lru_cache(maxsize=None)` also doesn't work.

seismiqb/src/horizon.py
```diff
@@ -976,14 +1027,164 @@ def get_cube_values(self, window=23, offset=0, scale=False, chunk_size=256):
...
+        >>> horizon.load_attribute('cube_values', (x_slice, i_slice, h_slice), window=10)
+
+        >>> horizon.load_attribute('heights')
+
+        >>> horizon.load_attribute('metrics', metrics='hilbert', normalize=False)
+
+        Notes
+        -----
+
+        Although the function can be used in a straightforward way as described above, originally it was implemented
```
SergeyTsimfer, line 1160: `Although ..., the main usage is...`
The story is nice, but not needed in the docstring

seismiqb/src/horizon.py
```diff
@@ -976,14 +1027,164 @@ def get_cube_values(self, window=23, offset=0, scale=False, chunk_size=256):
...
+        """
+        depths = [window // 2] if depths == 'mid' else depths
+        amplitudes = self.get_cube_values(window, **kwargs)
+        result = np.abs(hilbert(amplitudes))[:, :, depths]
+        result[self.full_matrix == self.FILL_VALUE] = np.nan
+        return self.normalize_by_binary_matrix(result, normalize)
+
+
+    @lru_cache(1)
+    def get_instantaneous_phase(self, window=23, depths=slice(None), normalize=False, use_cache=False, **kwargs):
```
SergeyTsimfer, line 1078: Feeling somewhat uneasy about the difference in default values for `depths` in methods

SergeyTsimfer: And `use_cache` is better be set to default `False` logic by using `apply_by_default` parameter of cache initialization

seismiqb/src/utils.py
```diff
@@ -104,6 +102,7 @@ def load_slide(cube_name, slide_no):
     Notes
     -----
     All arguments to the decorated method must be hashable.
+    Class defining decorated attribute must have `_cached_attributes` attribute, to allow tracking cached methods.
```
SergeyTsimfer, line 105: You can create this sequence yourself, if you need it.
As a side note: you can actually iterate over class attributes without evaluating any of the properties by using `isinstance(attr, property)` check.

SergeyTsimfer (approved):
I think it is time to better structure the code for the evaluation of geological attributes. This is to be done in a separate PR; as of now, I envision the `GeologicalAttributesMixin` class to contain all of those attributes with a more well-thought interface and re-usage of the same routines, as in `Metrics` (like phase evaluation).

seismiqb/src/utils.py
```diff
@@ -70,6 +70,19 @@ def stable_hash(key):
         key = key.encode('ascii')
     return str(blake2b(key).hexdigest())
 
+def flatten_nested(obj):
+    """ Recursively unwrap nested tuple, list or dictinary. """
```
SergeyTsimfer, line 74: typo
Add explanation of how various containers are flattened.
`obj` -> `iterable`

seismiqb/src/horizon.py
```diff
@@ -904,28 +914,73 @@ def add_to_mask(self, mask, locations=None, width=3, alpha=1, **kwargs):
...
-                heights += 1
-        return mask
+    def normalize_by_binary_matrix(self, arr, normalize='min-max'):
+        """ Normalize array only where `binary_matrix` is True.
 
+        Parameters
+        ----------
+        arr : np.array
+            Array to normalize.
+        normalize : str or dict
```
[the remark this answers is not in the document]
cdtn: Due to possible name collision. For example, some `Horizon` functions take `fill_value` argument. If weaved into its logic, `normalize_by_binary_matrix` will be forced to share the value of this argument with the outer function.

seismiqb/src/horizon.py
cdtn: > Feeling somewhat uneasy about the difference in default values for `depths` in methods

IMO, `None` default is even more counterintuitive than `'mid'`.

seismiqb/src/horizon.py
cdtn: I think, we still need it to explain why this function is implemented. Though, can be reworded a bit.

seismiqb/src/crop_batch.py
```diff
@@ -323,29 +327,81 @@ def load_cubes(self, ix, dst, src_locations='locations', src_geometry='geometrie
...
-            If -1 or 'all', all possible labels will be added.
-            If 1 or 'single', one random label will be added.
-            If array-like then elements are interpreted as indices of the desired labels
-            and must be ints in range [0, len(horizons) - 1].
-            Note if you want to pass an index of a single label it must be a list with one
-            element.
+            Dataset attribute with labels dict.
+        src_locations : str
+            Component of batch that stores locations of crops.
+        use_labels : str, int or sequence of ints
```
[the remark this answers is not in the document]
cdtn: <details> 
  <summary>1</summary>
  
</details>

<details> 
  <summary>2</summary>
  
</details>

<details> 
  <summary>3</summary>
  
</details>

seismiqb/src/horizon.py
```diff
@@ -919,13 +932,55 @@ def add_to_mask(self, mask, locations=None, width=3, alpha=1, **kwargs):
...
+                fill_value : any value compatible with `arr.dtype`
+                shift : number (for 'shift-rescale' mode)
+                rescale : number (for 'shift-rescale' mode)
+            Defaults to 'min-max'.
+        """
+        if not transform:
+            return arr
+
+        mask = self.put_on_full(self.binary_matrix, fill_value=False, dtype=bool)
+        values = arr[np.where(mask)]
```
SergeyTsimfer, line 961: Do you actually need `np.where` here?

seismiqb/src/horizon.py
```diff
@@ -919,13 +932,55 @@ def add_to_mask(self, mask, locations=None, width=3, alpha=1, **kwargs):
...
+        elif normalize == 'mean-std':
+            mean, std = values.mean(), values.std()
+            arr = (arr - mean) / std
+        elif normalize == 'shift-rescale':
+            arr = (arr + transform['shift']) * transform['rescale']
+        else:
+            raise ValueError('Unknown normalize mode {}'.format(normalize['normalize']))
+
+        fill_value = transform.get('fill_value', 0) if isinstance(transform, dict) else 0
+        arr[np.where(~mask)] = fill_value
```
SergeyTsimfer, line 978: `np.where` not needed

seismiqb/src/horizon.py
```diff
@@ -976,14 +1027,166 @@ def get_cube_values(self, window=23, offset=0, scale=False, chunk_size=256):
...
                 heights += 1
-
                 mask = heights < data_chunk.shape[2]
                 idx_i = idx_i[mask]
                 idx_x = idx_x[mask]
                 heights = heights[mask]
 
-        background[self.geometry.zero_traces == 1] = np.nan
-        return background
+        if nan_zero_traces:
```
SergeyTsimfer, line 1035: That (almost) can be done via `transform_by_binary_matrix`

seismiqb/src/horizon.py
```diff
@@ -919,13 +932,55 @@ def add_to_mask(self, mask, locations=None, width=3, alpha=1, **kwargs):
...
-            for _ in range(width):
-                mask[idx_i, idx_x, heights] = alpha
-                heights += 1
+            for shift in range(width):
+                mask[idx_i, idx_x, heights + shift] = alpha
         return mask
 
 
-    def get_cube_values(self, window=23, offset=0, scale=False, chunk_size=256):
+    def transform_by_binary_matrix(self, arr, transform='min-max'):
```
SergeyTsimfer, line 940: `transform_matrix`?
Normalize given by matrix where the horizon is

seismiqb/src/horizon.py
```diff
@@ -976,14 +1027,166 @@ def get_cube_values(self, window=23, offset=0, scale=False, chunk_size=256):
...
 
-        background[self.geometry.zero_traces == 1] = np.nan
-        return background
+        if nan_zero_traces:
+            background[self.geometry.zero_traces == 1] = np.nan
+        return self.transform_by_binary_matrix(background, transform)
+
+
+    @lru_cache(maxsize=1, apply_by_default=False)
+    def get_instantaneous_amplitude(self, window=23, depths=None, transform=False, **kwargs):
```
SergeyTsimfer, line 1041: `amplitudes`?

seismiqb/src/horizon.py
```diff
@@ -976,14 +1027,166 @@ def get_cube_values(self, window=23, offset=0, scale=False, chunk_size=256):
...
+            If slice or sequence of int, used for slicing calculated attribute along last axis.
+            If None, infer middle channel index from 'window' and slice at it calculated attribute along last axis.
+        transform : str or dict
+            For `Horizon.transform_by_binary_matrix`.
+        kwargs :
+            For `Horizon.get_cube_values`.
+
+        Notes
+        -----
+        Keep in mind, that hilbert transform produces artifacts at signal start and end. Therefore if you want to get
```
SergeyTsimfer, line 1090: `Hilbert`

seismiqb/src/horizon.py
```diff
@@ -976,14 +1027,166 @@ def get_cube_values(self, window=23, offset=0, scale=False, chunk_size=256):
...
+        """ Transform `matrix` attribute to match cubic coordinates.
+
+        Parameters
+        ----------
+        transform : str or dict
+            For `Horizon.transform_by_binary_matrix`.
+        kwargs :
+            For `put_on_full`.
+        """
+        matrix = self.put_on_full(**kwargs)
```
SergeyTsimfer, line 1113: Add the positional/kw argument of `what to transform`

seismiqb/src/utils.py
```diff
@@ -72,8 +70,21 @@ def stable_hash(key):
         key = key.encode('ascii')
     return str(blake2b(key).hexdigest())
 
+def tuplize_nested(iterable):
```
SergeyTsimfer, line 73: I believe that the `flatten` word can be helpful

seismiqb/src/utils.py
```diff
@@ -72,8 +70,21 @@ def stable_hash(key):
...
+    """ Recursively convert nested list or dictionary into nested tuple. """
+    result = []
+    if isinstance(iterable, (tuple, list)):
+        for item in iterable:
+            result.append(tuplize_nested(item))
+    elif isinstance(iterable, dict):
+        for key, value in sorted(iterable.items()):
+            result.append((key, tuplize_nested(value)))
```
SergeyTsimfer, line 81: `key` can be `tuple` too

seismiqb/src/cubeset.py
```diff
@@ -373,9 +431,9 @@ def show_slices(self, idx=0, src_sampler='sampler', n=10000, normalize=False, sh
     def show_points(self, idx=0, src_labels='labels', **kwargs):
         """ Plot 2D map of points. """
         map_ = np.zeros(self.geometries[idx].cube_shape[:-1])
-        for label in getattr(self, src_labels)[idx]:
+        for label in self[idx, src_labels]:
             map_[label.points[:, 0], label.points[:, 1]] += 1
-        labels_class = type(getattr(self, src_labels)[idx][0]).__name__
+        labels_class = type(self[idx, src_labels, 0]).__name__
```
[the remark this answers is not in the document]
cdtn: https://github.com/gazprom-neft/seismiqb/issues/17#issuecomment-744354533

seismiqb/src/cubeset.py
```diff
@@ -132,29 +167,161 @@ def create_labels(self, paths=None, filter_zeros=True, dst='labels', labels_clas
...
+            label_list = [labels_class(path, self.geometries[idx], **kwargs) for path in paths[idx]]
             label_list.sort(key=lambda label: label.h_mean)
             if filter_zeros:
                 _ = [getattr(item, 'filter')() for item in label_list]
-            getattr(self, dst)[ix] = [item for item in label_list if len(item.points) > 0]
+            self[idx, dst] = [item for item in label_list if len(item.points) > 0]
+            self._cached_attributes.add(dst)
+
+
+    def show_labels(self, indices=None, main_labels='labels', overlay_labels=None, attributes=None, correspondence=None,
```
SergeyTsimfer, line 199: Let's leave it for a future PR

SergeyTsimfer (commented):
Makes me think about a separate child of `Cubeset` to avoid clutter in the current class

seismiqb/src/cubeset.py
cdtn: Let's keep it here and fix later if needed.

seismiqb/src/horizon.py
```diff
@@ -1165,7 +1368,12 @@ def filled_matrix(self):
     @property
     def full_matrix(self):
         """ Matrix in cubic coordinate system. """
-        return self.put_on_full()
+        return self.get_full_matrix()
+
+    @property
+    def presence_matrix(self):
+        """ Binary matrix in cubic coordinate system. """
```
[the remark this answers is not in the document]
cdtn: So as `Horizon.full_matrix` doc.

seismiqb/src/horizon.py
```diff
@@ -919,13 +930,53 @@ def add_to_mask(self, mask, locations=None, width=3, alpha=1, **kwargs):
...
+        """ Normalize array where horizon is present, fill in with constant where it is not.
+
+        Parameters
+        ----------
+        arr : np.array
+            Array to transform. Its shape at first two dimensions must match `self.presence_matrix` shape.
+        normalize : 'min-max', 'mean-std', 'shift-rescale' or False
+            Normalization mode for data where `presence_matrix` is True.
+            If None, no normalization applied. Defaults to False.
+        fill_value : any value compatible with `arr.dtype`
```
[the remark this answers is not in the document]
cdtn: Well, it can be `np.unicode_` one day... Is it a `number`?

seismiqb/src/horizon.py
```diff
@@ -919,13 +930,53 @@ def add_to_mask(self, mask, locations=None, width=3, alpha=1, **kwargs):
...
 
-    def get_cube_values(self, window=23, offset=0, scale=False, chunk_size=256):
+    def transform_where_present(self, arr, normalize=None, fill_value=None, shift=None, rescale=None):
+        """ Normalize array where horizon is present, fill in with constant where it is not.
+
+        Parameters
+        ----------
+        arr : np.array
+            Array to transform. Its shape at first two dimensions must match `self.presence_matrix` shape.
+        normalize : 'min-max', 'mean-std', 'shift-rescale' or False
```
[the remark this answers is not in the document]
cdtn: Comment purpose not understood. If you think `None` is a better option than `False`, just say it.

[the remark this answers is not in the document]
SergeyTsimfer: The syntax of docstring is different

seismiqb/src/utils.py
```diff
@@ -15,9 +16,10 @@
 
 
 
-def file_print(msg, path):
+# pylint: disable=redefined-outer-name
+def file_print(msg, path, mode='w'):
     """ Print to file. """
-    with open(path, 'w') as file:
+    with open(path, mode) as file:
```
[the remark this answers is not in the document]
cdtn: This is outdated, like, three commits ago.

seismiqb/src/utils.py
```diff
@@ -760,3 +786,11 @@ def local_correlation(region):
             if den != 0:
                 corr[i, j] = cov / den
     return np.mean(corr)
+
+def pop_kwargs_for_function(func, **kwargs):
+    """ Pop variables from kwargs that are meant for function. """
```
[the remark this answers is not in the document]
cdtn: > Signature should be `function, dictionary`: no unpacking here (and at the calls)

Comment is outdated.

seismiqb/src/utils.py
```diff
@@ -72,8 +72,21 @@ def stable_hash(key):
         key = key.encode('ascii')
     return str(blake2b(key).hexdigest())
 
+def to_nested_tuple(iterable):
```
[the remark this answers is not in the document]
cdtn: This will make possible future inspection of cache keys more puzzling.

seismiqb/src/horizon.py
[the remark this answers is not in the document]
SergeyTsimfer: If it is to become anything but a `int`/`float`, which are `number`, we would have much bigger problems to worry about

seismiqb/src/utils.py
```diff
@@ -95,15 +108,21 @@ class lru_cache:
 
     Examples
     --------
-    Store loaded slides::
+    Store loaded slides by default:
```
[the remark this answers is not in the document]
cdtn: Why?

SergeyTsimfer (approved):

SergeyTsimfer (approved):
Somewhat useful PR: merging

PR #49, opened 2020-12-15
Fix req

SergeyTsimfer:


roman-kh (commented):
Add an action which check that reqs are correct (by installing reqs into an empty env and importing the library and the running some easy tests)

PR #53, opened 2020-12-29
Plotter update

akoryagin:
## Main additions

- split `convert_kwargs` into methods into each plotter-class
- add `grid` plot-mode
- add demo-nb with examples


seismiqb/src/plotters.py
```diff
@@ -125,6 +76,37 @@ def plot_loss(*data, title=None, **kwargs):
 class MatplotlibPlotter:
```
SergeyTsimfer, line 76: `self` is never used in the methods of `MatplotlibPlotter` (or in the `PlotlyPlotter`, for that matter). Those classes are used to be namespaces for a group of functions. Make them proper namespaces by converting methods to `classmethods` and getting rid of those `self` in every signature.

Also, that would allow not to create an instance of that class to call its methods

seismiqb/src/plotters.py
```diff
@@ -125,6 +76,37 @@ def plot_loss(*data, title=None, **kwargs):
...
+            **keys_converter,
+            'zmin': 'vmin', 'zmax': 'vmax',
+            'xaxis': 'xlabel', 'yaxis': 'ylabel'
+        }
+        # make conversion-procedure for key-value pairs
+        def converter(k, v):
+            if k in ('xaxis', 'yaxis'):
+                return keys_converter[k], v['title_text']
+            return keys_converter[k], v
+
```
SergeyTsimfer, line 102:
```python
converted = {}
for key, value in kwargs.items():
    if key in keys_converter:
        new_key = keys_converter[key]
        if key in ['xaxis', 'yaxis']:
             converted[new_key] = value.get('title_text', '')
        else:
            converted[new_key] = value
    else:
        converted[key] = value
```

seismiqb/src/plotters.py
```diff
@@ -331,6 +313,81 @@ def wiggle(self, image, **kwargs):
         self.save_and_show(fig, **updated)
 
 
+    def grid(self, images, single_kwargs=None, **kwargs):
+        """ Make grid of plots using range of images and info about how the grid should be organized.
```
SergeyTsimfer, line 317: Most of the docstrings are unreachable anyway; make them short -- no need to repeat the list of possible args for `mpl`

seismiqb/src/plotters.py
```diff
@@ -331,6 +313,81 @@ def wiggle(self, image, **kwargs):
         self.save_and_show(fig, **updated)
 
 
+    def grid(self, images, single_kwargs=None, **kwargs):
```
SergeyTsimfer, line 316: What about the following signature:
```python
def grid(sequence, **kwargs)
```
- `**kwargs` are common parameters for each of the plots
- `sequence` can be a list of images
- `sequence` can be a list of dicts: each of them contains image, as well as parameters for its display. Those parameters take precedence over the `**kwargs`
- all of the subplot parameters are taken exactly the same way you do now

PR #67, opened 2021-02-26
Controller refactor

SergeyTsimfer:
This PR rewrites `BaseController` and derived classes anew. Among other changes, it proposes to:

- use single `config`-dictionary to parameterize the entire run of the controller
- explicitly pass information between methods
- use instances only to hold parameters
- log pretty much every step of `train` and `inference`, with both textual (iteration-wise) info and graphs (resource monitors and horizon maps)

Changes to come:
- [x] modify derived classes (`Interpolator`, `Enhancer`, `Extender`)
- [ ]  improve `evaluate` method
- [x]  update docstrings
- [x]  remove old version


pylintrc
```diff
@@ -17,7 +17,7 @@ disable=no-value-for-parameter, no-self-use, too-few-public-methods, unsubscript
         no-member, arguments-differ, len-as-condition, keyword-arg-before-vararg, too-many-locals, too-many-lines,
         logging-format-interpolation, no-name-in-module, import-error, cyclic-import, duplicate-code, import-self,
         abstract-method, relative-beyond-top-level, unused-argument, too-many-public-methods, too-many-instance-attributes,
-        invalid-name
+        invalid-name, consider-using-generator
```
cdtn, line 20: Why disable it globally?

seismiqb/src/controllers/new.py
```diff
@@ -0,0 +1,677 @@
...
+            'device': 'gpu',
+
+            'add_prefix': False,
+            'dump': False,
+            'name': '',
+        }
+    })
+
+    def __init__(self, config):
+        self.config = Config(copy(self.DEFAULTS))
```
cdtn, line 84: Why copy?

seismiqb/src/controllers/new.py
```diff
@@ -0,0 +1,677 @@
...
+
+        # Log: resource graphs
+        if self.monitor:
+            monitor.__exit__(None, None, None)
+            monitor.visualize(savepath=self.make_savepath('末 train_resource.png'), show=self.plot)
+
+        # Log: loss over iteration
+        plot_loss(model.loss_list, show=self.plot,
+                  savepath=self.make_savepath('末 model_loss.png'))
+        finish_loss = np.mean(model.loss_list[-25:])
```
cdtn, line 327: Finishing/final?

seismiqb/src/controllers/new.py
```diff
@@ -0,0 +1,677 @@
...
+
+    # Evaluate
+    def evaluate(self, predictions, targets=None, dataset=None, **kwargs):
+        """ !!. """
+        #pylint: disable=cell-var-from-loop
+        config = Config({**self.config['evaluate'], **kwargs})
+        add_prefix, dump, name = config.pop(['add_prefix', 'dump', 'name'])
+        supports, device = config.pop(['supports', 'device'])
+
+
```
cdtn, line 508:
```suggestion

```

seismiqb/src/controllers/new.py
```diff
@@ -0,0 +1,677 @@
...
+                                plot=True, show=self.plot,
+                                savepath=self.make_savepath(*prefix, name + 'instantaneous_phase.png'))
+
+            perturbed_mean, perturbed_max = hm.evaluate('perturbed', device=device,
+                                                        plot=True, show=self.plot,
+                                                        savepath=self.make_savepath(*prefix, name + 'perturbed.png'))
+
+            # Compare to targets
+            if targets:
+                _, oinfo = hm.evaluate('find_best_match', agg=None)
```
cdtn, line 543: o?

seismiqb/src/controllers/new.py
```diff
@@ -0,0 +1,677 @@
...
+
+    def log_to_file(self, msg, path):
+        """ Log message to a separate file. """
+        log_path = self.make_savepath(path)
+        if log_path:
+            with open(log_path, 'w') as file:
+                print(msg, file=file)
+
+    # Dataset creation: geometries, labels, grids, samplers
+    def make_dataset(self, cube_paths=None, horizon_paths=None, horizon=None):
```
AlexeyKozhevin, line 146: Let's extend it to use it not only with horizons.

AlexeyKozhevin: At least add `labels_class`

seismiqb/src/controllers/new.py
```diff
@@ -0,0 +1,677 @@
...
+        """
+        if horizon is not None:
+            dataset = SeismicCubeset.from_horizon(horizon)
+        else:
+            dataset = SeismicCubeset(cube_paths)
+            dataset.load_geometries()
+
+            if horizon_paths:
+                if isinstance(horizon_paths, str):
+                    horizon_paths = {dataset.indices[0]: glob(horizon_paths)}
```
AlexeyKozhevin, line 174: What if I want several cubes in the dataset but with the same relative path to labels?

seismiqb/src/controllers/new.py
```diff
@@ -0,0 +1,677 @@
...
+        # Start resource tracking
+        if self.monitor:
+            monitor = Monitor(['uss', 'gpu', 'gpu_memory'], frequency=0.05, gpu_list=self.gpu_list)
+            monitor.__enter__()
+
+        # Make pipeline
+        train_pipeline = self.get_train_template(**kwargs) << pipeline_config << dataset
+
+        # Log: pipeline_config to a file
+        self.log_to_file(pformat(pipeline_config.config, depth=2), '末 train_config.txt')
```
AlexeyKozhevin, line 285: 末?

AlexeyKozhevin (changes requested):

seismiqb/src/controllers/base.py
```diff
@@ -1,728 +1,275 @@
...
 
-    def make_save_path(self, *postfix):
-        """ Create nested path from provided strings; create corresponding directories.
+        gpu_list = literal_eval(os.getenv('CUDA_VISIBLE_DEVICES'))
+        self.gpu_list = list(gpu_list) if isinstance(gpu_list, tuple) else [gpu_list]
+        self.make_filelogger()
 
-        If `save_dir` attribute is None, then None is returned: that is used as signal to omit saving
+    # Utility functions
+    def make_savepath(self, *postfix):
```
cdtn, line 98:
```suggestion
    def make_savepath(self, *subdirs):
```

cdtn (changes requested):
Also:
1. Refactor `01_DEMO_*` notebooks to have only predict pipeline.
2. Replace `model_config` insides with `…` while logging in `BaseController`.

seismiqb/src/cubeset.py
```diff
@@ -1178,14 +1178,13 @@ def assemble_crops(self, crops, grid_info='grid_info', order=None, fill_value=0)
         # Do nothing if number of crops differ from number of points in the grid.
         if len(crops) != len(grid_info['grid_array']):
             raise ValueError('Length of crops must be equal to number of crops in a grid')
-        order = order or (2, 0, 1)
-        crops = np.array(crops)
-        if len(crops) != 0:
+
+        if fill_value is None and len(crops) != 0:
             fill_value = np.min(crops)
```
AlexeyKozhevin, line 1183: nanmin

AlexeyKozhevin: and also `background[background_slice] = np.maximum(crop, previous)` -> `background[background_slice] = np.nanmax([crop, previous], axis=0)`

SergeyTsimfer: `maximum` is element-wise; we can use `fmax` to ignore `NaN`s, but it is much slower:(

seismiqb/src/controllers/base.py
```diff
@@ -1,728 +1,275 @@
...
+        Model instance
         """
-        cube_path = horizon.geometry.path
-
-        dsi = FilesIndex(path=[cube_path], no_ext=True)
-        dataset = SeismicCubeset(dsi)
-        dataset.geometries[dataset.indices[0]] = horizon.geometry
-        dataset.labels[dataset.indices[0]] = [horizon]
+        # Prepare parameters
+        pipeline_config = Config({**self.config['train'], **kwargs})
```
AlexeyKozhevin, line 165: Add `common` level to `DEFAULT_CONFIG` to avoid keys duplication in `train` and `inference`

seismiqb/src/cubeset.py
AlexeyKozhevin: So what do you propose to work with crops which possibly can have nans?

seismiqb/src/controllers/base.py
```diff
@@ -1,728 +1,275 @@
...
 
-        msg = '\n'
-        for idx in dataset.indices:
-            msg += f'{idx}\n'
-            for hor in dataset.labels[idx]:
-                msg += f'    {hor.name}'
-        self.log(f'Created dataset ::: {msg}')
-        return dataset
+    # Train
+    def train(self, dataset, **kwargs):
```
AlexeyKozhevin, line 155: It will be nice to add parameter `config` to substitute `config` directly to methods. In that case I can use the controller in Research in the following way: define controller instance before research creation and then add callables from it (`make_dataset`, `train`, ...) to research. If each callable will have `config` parameter then I will have an opportunity to substitute config from domain into callable by `RC()`.

AlexeyKozhevin: I also think that it will be a good opportunity to add classes to research. Instances of them will be created at the first iteration and then you will call methods of them

seismiqb/src/controllers/base.py
SergeyTsimfer: The idea is to configure each of them completely separate, as explicitly as possible

seismiqb/src/controllers/base.py
SergeyTsimfer: Added the `config` parameter for train/inference methods

PR #68, opened 2021-03-19
Add geometry name and path anonymization

cdtn:
Use this statement in your jupyter notebooks to mask field name in publicly displayed geometry path and name.

```
%env SEISMIQB_ANONYMIZE 1
```


SergeyTsimfer (approved):
Very useful PR!

PR #77, opened 2021-05-12
Online crops aggregation

a-arefina:
the difference with current aggregation is that in current version fill value for areas not covered with crops (ex. zero traces) is inferred from all crops, but in this online mode it is -np.inf


models/Benchmarks.ipynb
```diff
@@ -752,14 +753,70 @@
    "source": [
     "%%time\n",
     "dataset.make_grid(dataset.indices[0], CROP_SHAPE,\n",
-    "                  [0, 417], [0, 868], [800, 1000],\n",
-    "                  overlap=(1, 96, 96),\n",
+    "                  [0, 400], [0, 868], [800, 1000],\n",
+    "                  overlap=(0, 96, 96),\n",
```
SergeyTsimfer, line 757: It is better to change `overlap` to `strides` and keep the same value `(1, 96, 96)`

seismiqb/src/crop_batch.py
```diff
@@ -1244,3 +1244,109 @@ def plot_components(self, *components, idx=0, slide=None, mode='overlap', order_
...
+        order : tuple
+            The order of axes of the crop which corresponds to natural iline-xline-depth order
+        """
+        crop = self.get(ix, src)
+        location = self.get(ix, src_locations)
+        container.put(crop.transpose(order), tuple(location))
+
+class BaseAggregationContainer:
+    """ Container for on-line aggregation of crops """
+    def __init__(self, shape=None, ilines=None, xlines=None, heights=None):
```
SergeyTsimfer, line 1269: As far as I can see, the `*Container` is always used in a combination with `make_grid` method. It would be easier to create it as one of the attributes of `Dataset` right in the `make_grid` call. That would require lazy initialization of underlying storages (`self.data` and `self.counters` in case of `avg` container, for example), but it is for the better as well imo

seismiqb/src/crop_batch.py
```diff
@@ -1244,3 +1244,109 @@ def plot_components(self, *components, idx=0, slide=None, mode='overlap', order_
...
+    def aggregate(self):
+        """ Computes and returns aggregated cube.
+        Can leave the container in dirty state, so it should be called only once """
+        self.dirty = True
+        return self._aggregate()
+
+    def _aggregate(self):
+        raise NotImplementedError
+
+class AvgContainer(BaseAggregationContainer):
```
SergeyTsimfer, line 1322: Should we add the `reset` method? If the lazy initialization of storages (like `self.data`) is implemented, that would be very easy (just setting redundant to `None`), and would ease the memory management. Otherwise, for example, in the  `Benchmark` notebook you keep those unnecessary storages in memory

seismiqb/src/crop_batch.py
```diff
@@ -1244,3 +1244,109 @@ def plot_components(self, *components, idx=0, slide=None, mode='overlap', order_
...
+
+    def _aggregate(self):
+        self.counters[self.counters == 0] = 1
+        self.data /= self.counters
+        return self.data
+
+class MaxContainer(BaseAggregationContainer):
+    """ Maximum aggregation of crops """
+
+    FILL_VAL = -np.inf
```
SergeyTsimfer, line 1341: `FILL_VALUE`, to be consistent with `seismiqb`. Also, I would use the same names, as `seismiqb.src.utility_classes.Accumulator` class, as the `Container` essentially does the same.

SergeyTsimfer: And those new classes need to be moved in `utility_classes.py`

seismiqb/src/crop_batch.py
```diff
@@ -1244,3 +1244,109 @@ def plot_components(self, *components, idx=0, slide=None, mode='overlap', order_
         }
 
         plot_image(imgs, mode=mode, order_axes=order_axes, **kwargs)
+
+    @action
+    @inbatch_parallel(init='indices', post='_assemble', target='for')
+    def dump(self, ix, src, container, src_locations='locations', order=(1, 2, 0)):
```
SergeyTsimfer, line 1250: `update_container`?
Also, make the default `order=(0, 1, 2)`

seismiqb/src/crop_batch.py
```diff
@@ -1244,3 +1244,109 @@ def plot_components(self, *components, idx=0, slide=None, mode='overlap', order_
...
+        container : BaseAggregationContainer
+            Container for resulting cube aggregation.
+        src_locations : src
+            Component with crop location, default: locations
+        order : tuple
+            The order of axes of the crop which corresponds to natural iline-xline-depth order
+        """
+        crop = self.get(ix, src)
+        location = self.get(ix, src_locations)
+        container.put(crop.transpose(order), tuple(location))
```
SergeyTsimfer, line 1265: What is the reason for conversion to `tuple`? You use it only to iterate over

seismiqb/src/crop_batch.py
```diff
@@ -1244,3 +1244,23 @@ def plot_components(self, *components, idx=0, slide=None, mode='overlap', order_
...
+        container : BaseAggregationContainer
+            Container for resulting cube aggregation.
+        src_locations : src
+            Component with crop location, default: locations
+        order : tuple
+            The order of axes of the crop which corresponds to natural iline-xline-depth order
+        """
+        crop = self.get(ix, src)
+        location = self.get(ix, src_locations)
+        container.put(crop.transpose(order), location)
```
SergeyTsimfer, line 1265: `return self`

PR #79, opened 2021-05-14
fix hdf5 cube_shape for non-default projections

a-arefina:


SergeyTsimfer (approved):

seismiqb/src/geometry/segy.py
```diff
@@ -601,13 +605,18 @@ def convert(self, format='blosc', path=None, postfix='', projections='ixh',
...
-                for idx in range(self.cube_shape[axis]):
-                    slide = self.load_slide(idx, axis=axis, stable=False)
-                    slide = slide.T if axis == 1 else slide
+                for idx in range(self.cube_shape[load_axis]):
+                    slide = self.load_slide(idx, axis=load_axis, stable=False)
+                    slide = slide if axis == 0 else slide.T
                     slide = transform(slide)
 
-                    cube[idx, :, :] = slide
+                    if axis == 2:
```
SergeyTsimfer, line 616: The problem with that approach is that `blosc` slided storage would not allow this

a-arefina: Should we add an 'if hdf5' condition? I've already discussed this issue, but haven't thought of anything better. Without this fix neither of hdf5 and blosc is created

a-arefina: If memory affords this (and, as both our old and new assembling procedures demonstrate, it does) we could load all cube into memory iline-wise and then dump it to any projection in any format we choose

SergeyTsimfer (approved):

PR #82, opened 2021-06-15
Updates in model input design

HollowPrincess:
- [x] It needs an approvement of https://github.com/analysiscenter/batchflow/pull/592
- [x] Check run after updates in the previous PR


seismiqb/src/controllers/best_practices.py
```diff
@@ -1,6 +1,17 @@
...
+    def __init__(self, width=3):
+        super().__init__()
+        self.width_weights = torch.ones((1, 1, 1, width))
+        
+    @torch.cuda.amp.autocast(enabled=False)
+    def forward(self, x):
+        x = torch.nn.functional.softmax(x, dim=-1)
+        width_weights = self.width_weights.to(device=x.device, dtype=x.dtype)
+        x = F.conv2d(x, width_weights, padding=(0, 1))
+        return x.float()
```
HollowPrincess, line 14: I am not sure that this class should be in this code file.
Also, there is a problem with the usability: in a config (e.g., line 84), we pass the class name, and we can't pass the ```width``` argument because we can't pass an instance in a config. (When we pass an instance of activation function in a model config, it has an extra addition at the beginning of a layer.)

seismiqb/src/controllers/best_practices.py
```diff
@@ -1,6 +1,23 @@
 """ Collection of good architectures for the tasks of horizon detection. """
+import torch
+from torch import nn
+from torch import functional as F
+
 from ...batchflow.batchflow.models.torch import ResBlock
+from ...batchflow.models.torch.losses.binary import Dice
+
 
+class DepthSoftmax(nn.Module):
```
SergeyTsimfer, line 10: Docstring

seismiqb/src/crop_batch.py
```diff
@@ -619,23 +620,27 @@ def concat_components(self, src, dst, axis=-1):
         axis : int
             The axis along which the arrays will be joined.
         """
-        if axis != -1:
-            raise NotImplementedError("For now function works for `axis=-1` only.")
-        _ = dst
+        if not isinstance(src, (list, tuple, np.ndarray)):
+            raise ValueError()
```
SergeyTsimfer, line 624: Either provide a useful error message, or don't raise anything
What would be the error without that if-clause guard?

SergeyTsimfer (approved):
Please, merge master branch

seismiqb/src/controllers/best_practices.py
HollowPrincess: Another question is about passing activation parameters into model config

PR #86, opened 2021-06-16
Add Accumulator3D

SergeyTsimfer:
This PR improves on @a-arefina work by allowing us to accumulate predicted crops either in memory or on disk, while also adding more aggregations and QoL changes to placeholder creation.

I will also include the following changes as soon as possible:

- [x] change `HorizonController` to use new accumulate logic
- [ ] add non-zero `prefetch` value for inference of `HorizonController`
- [x] correct `SeismicCubeset.show_slices` with `mask_rebatch`
- [x] augment `SeismicGeometry` repr with the size of original `SEG-Y` cube and area of the processed field

### DIFF
--- a/seismiqb/src/utility_classes.py
+++ b/seismiqb/src/utility_classes.py
@@ -1,4 +1,5 @@
 """ Helper classes. """
+import os
 from time import perf_counter
 from collections import OrderedDict, defaultdict
 from threading import RLock
@@ -14,6 +15,7 @@ except ImportError:
     cp = np
     CUPY_AVAILABLE = False
 from numba import njit
+import h5py
 
 from ..batchflow import Sampler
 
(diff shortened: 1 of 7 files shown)


SergeyTsimfer:
As of now, `HorizonController.inference` correctly works with non-zero prefetch, but leaks memory

seismiqb/src/utility_classes.py
```diff
@@ -224,6 +226,239 @@ def get(self, final=False):
...
+            self.options = {**kwargs}
+        self.type = 'hdf5' if path is not None else 'numpy'
+
+        self.names = []
+        self.aggregated = False
+
+        # Create underlying storages
+        self._init()
+
+    def _init(self):
```
a-arefina, line 293: why did you remove __init__ methods in child classes? You could pass specialized arguments there, instead of mixing them all in base classes __init__ or parsing kwargs

a-arefina (commented):
still reflecting on placeholders, but here are some comments

seismiqb/src/utility_classes.py
```diff
@@ -451,11 +451,15 @@ def _aggregate(self):
...
-                self.data[i] = np.pow(self.data[i], 1/counts)
+
+                counts = counts.astype(np.float32)
+                counts **= -1
+                self.data[i] **= counts
 
         elif self.type == 'numpy':
             self.counts[self.counts == 0] = 1
-            self.data = np.pow(self.data, 1/self.counts)
+            self.counts **= -1
```
a-arefina, line 461: add 
counts = counts.astype(np.float32)

seismiqb/src/utility_classes.py
```diff
@@ -224,6 +226,247 @@ def get(self, final=False):
...
+        _ = crop, location
+        raise NotImplementedError
+
+    def aggregate(self):
+        """ Finalize underlying storages to create required aggregation. """
+        if self.aggregated:
+            raise RuntimeError('All data in the container has already been cleared!')
+        self._aggregate()
+
+        # Re-open the HDF5 file to optionally repack it
```
a-arefina, line 339: plz update the comment to describe why we need to reopen the file

seismiqb/src/utility_classes.py
```diff
@@ -224,6 +226,247 @@ def get(self, final=False):
...
+        After the initialization, we keep the file handle in `w-` mode during the update phase.
+        After aggregation, we re-open the file to automatically repack it in `r` mode.
+    kwargs : dict
+        Other parameters are passed to HDF5 dataset creation.
+    """
+    def __init__(self, shape=None, origin=None, dtype=np.float32, transform=None, path=None, **kwargs):
+        # Dimensionality and location
+        self.shape = shape
+        self.origin = origin
+
```
a-arefina, line 270: maybe add self.data = None to show that this field must exist?

a-arefina (approved):
please see two minor comment and off we go!

PR #102, opened 2021-07-14
Add flatten method to IndexedDict

cdtn:


seismiqb/src/utility_classes.py
```diff
@@ -496,6 +503,21 @@ def __setitem__(self, key, value):
             key = list(self.keys())[key]
         super().__setitem__(key, value)
 
+    def flatten(self, keys=None):
+        """ Get dict values for requested keys in a single list. """
+        all_keys = list(self.keys())
+        keys = to_list(keys, default=all_keys, dtype='object')
```
SergeyTsimfer, line 509: Why is that not `to_list(keys if keys is not None else all_keys, dtype='object')`?

SergeyTsimfer: And `all_keys` are already `list`, and it was made with built-in

SergeyTsimfer (approved):

SergeyTsimfer (approved):

PR #117, opened 2021-08-02
Fix unprocessed geometry repr

cdtn:


seismiqb/src/geometry/base.py
```diff
@@ -578,7 +578,10 @@ def ngbytes(self):
 
     # Textual representation
     def __repr__(self):
-        return f'<Inferred geometry for cube {self.displayed_name}: {tuple(self.cube_shape)}>'
+        geometry = f'geometry for cube {self.displayed_name}'
```
SergeyTsimfer, line 581: `msg`
Also, add the same for `__str__`

SergeyTsimfer (approved):

PR #123, opened 2021-08-09
Labels minor improvements

cdtn:
Add new methods

1. `Facies.cut_subset` — creates a new label by cutting out a subset from the initial label defined by the provided matrix.
2. `VisualizationMixin.make_savepath` — replaces an asterisk with label name in the provided path and creates save directory if it doesn't already exist.


seismiqb/src/labels/horizon_visualization.py
```diff
@@ -355,3 +362,15 @@ def make_triangulation(self, n_points, threshold, slices, **kwargs):
         simplices = filter_simplices(simplices=tri.simplices, points=tri.points,
                                      matrix=self.full_matrix, threshold=threshold)
         return x, y, z, simplices
+
+    # Supplementary
+    def make_savepath(self, path, extension=''):
```
SergeyTsimfer, line 367: Should we move it in `utils`, with signature `def make_savepath(path, extension, name)`?

SergeyTsimfer (approved):

seismiqb/src/labels/horizon_facies.py
```diff
@@ -74,6 +74,29 @@ def get_subset(self, name):
...
+            Name for new label. If None, original label name used.
+
+        Returns
+        -------
+        New `Facies` instance with filtered points matrix.
+        """
+        other = copy(self)
+        other.name = name or self.name
+
+        filtering_matrix = (matrix < 0.5).astype(int)
```
SergeyTsimfer, line 96: I would rather pass already thresholded matrix into this function

SergeyTsimfer (approved):

PR #142, opened 2021-10-15
Faults improvements

AlexeyKozhevin:
- Refactor faults controller: remove several methods for inference and create only one
- Add new skeletonize procedure
- Add smoothing procedure of predictions
- `make_segy_from_array` can create segy with specs from other file: is needed to infer the part of the cube and save into segy


seismiqb/src/labels/fault.py
```diff
@@ -182,6 +196,21 @@ def add_to_mask(self, mask, locations=None, **kwargs):
...
+            indices = np.isin(points[:, self.direction], slides_indices)
+            points = points[indices]
+            mask_pos = np.isin(
+                np.arange(mask.shape[self.direction]),
+                slides_indices - locations[self.direction].start
+            )
+            if mask_pos.any():
+                if self.direction == 0:
+                    # mask[mask_pos] = 0
+                    mask[mask_pos] = np.clip(mask[mask_pos], 0, 1)
```
SergeyTsimfer, line 210: `mask[mask_pos, :] = np.clip(mask[mask_pos, :], 0, 1)`

seismiqb/src/labels/fault.py
```diff
@@ -219,6 +248,15 @@ def split_file(cls, path, dst):
         df = pd.read_csv(path, sep=r'\s+', names=cls.FAULT_STICKS)
         df.groupby('name').apply(cls.fault_to_csv, dst=dst)
 
+    def merge(self, other, **kwargs):
+        """ Merge two Fault instances"""
+        points = np.concatenate([self.points, other.points], axis=0)
```
SergeyTsimfer, line 253: In each separate Fault instance `points` (and `nodes`) are sorted along the first two columns. Breaking this can cause troubles in future (or even now)
Also, I do have an efficient `njitted` function to merge two such sorted arrays into a new sorted `points`
😉

AlexeyKozhevin: And where is it?

SergeyTsimfer: Something like that should work:

```python
from numba import njit

@njit
def concat_sorted(first_array, second_array):
    buffer = np.zeros((len(first_array) + len(second_array), 3), dtype=np.int32)
    c = 0
    i, j = 0, 0


    # Condition is re-evaluated to swapped arrays
    while i < len(first_array) and j < len(second_array):
        first_ = first_array[i]
        second_ = second_array[j]

        # Need to swap or not
        if second_[0] == first_[0]:
            if second_[1] < first_[1]:
                first_array, second_array = second_array, first_array
                i, j = j, i
            elif second_[1] == first_[1]:
                # Same value: no need to duplicate
                j += 1

        elif second_[0] < first_[0]:
            first_array, second_array = second_array, first_array
            i, j = j, i

        buffer[c] = first_array[i]
        c += 1

        i += 1

    return buffer[:c]
```

seismiqb/src/labels/fault_postprocessing.py
```diff
@@ -0,0 +1,162 @@
+""" Utils for faults postprocessing. """
+
+import numpy as np
+from numba import njit, prange
+from numba.types import bool_
+
+@njit(parallel=True)
+def skeletonize(slide, width=5, rel_height=0.5, threshold=0.05):
```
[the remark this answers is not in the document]
AlexeyKozhevin: It applies `find_peaks` to multiple arrays in parallel

seismiqb/src/labels/fault_postprocessing.py
```diff
@@ -0,0 +1,162 @@
...
+        is for labels
+    Returns
+    -------
+    sizes : numpy.ndarray
+    """
+    sizes = []
+    for array in labels:
+        i_len = (array[:, 0].max() - array[:, 0].min())
+        x_len = (array[:, 1].max() - array[:, 1].min())
+        sizes += [(i_len ** 2 + x_len ** 2) ** 0.5]
```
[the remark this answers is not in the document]
AlexeyKozhevin: There is no big difference for us: at the current stage, we don't use sizes. They were needed to sort faults when we tried to split them into separate objects

PR #143, opened 2021-10-19
Adjust default figsize calculation

cdtn:
Infer figure size from aspect ratios of provided data.

Also, create a basic colormap from the 'ocean' one.


SergeyTsimfer (approved):

seismiqb/src/field/visualization.py
```diff
@@ -357,100 +217,166 @@ def show(self, attributes='snr', mode='imshow', return_figure=False, short_title
...
+        return load_params
+
+    @staticmethod
+    def _make_cmap(params):
+        linkage = {
+            'Depths': ['depths', 'matrix', 'full_matrix'],
+            'Reds': ['spikes', 'quality_map'],
+            'Metric': ['metric', 'metrics'],
+            'generate': ['masks', 'full_binary_matrix']
+        }
```
HollowPrincess, line 353: Why ```'generate'``` starts with a lower case letter and other points are not?

seismiqb/src/field/visualization.py
```diff
@@ -357,100 +217,166 @@ def show(self, attributes='snr', mode='imshow', return_figure=False, short_title
...
+
+        attribute_name = params['attribute_name']
+        for cmap, names in linkage.items():
+            if attribute_name in names:
+                if cmap == 'generate':
+                    global_name = ''.join(filter(lambda x: x.isalpha(), attribute_name))
+                    if global_name not in NAME_TO_COLOR:
+                        NAME_TO_COLOR[global_name] = next(COLOR_GENERATOR)
+                    cmap = NAME_TO_COLOR[global_name]
+                return cmap
```
HollowPrincess, line 363: Maybe: ```if attribute_name in names and cmap == 'generate'```?

seismiqb/src/field/visualization.py
```diff
@@ -357,100 +217,166 @@ def show(self, attributes='snr', mode='imshow', return_figure=False, short_title
...
+
+        return 'Basic'
+
+    @staticmethod
+    def _make_alpha(params):
+        attribute_name = params['attribute_name']
+        if attribute_name in ['masks', 'full_binary_matrix']:
+            return 0.7
```
HollowPrincess, line 371: Why 0.7?

seismiqb/src/field/visualization.py
```diff
@@ -357,100 +217,166 @@ def show(self, attributes='snr', mode='imshow', return_figure=False, short_title
...
+        return 1.0
+
+    @staticmethod
+    def _make_title(params):
+        label_name = params['label_name']
+        attribute_name = params['attribute_name']
```
HollowPrincess, line 377: Maybe ```params.get(smth)``` will be more safely? Or this is unimportant for this case?

seismiqb/src/field/visualization.py
```diff
@@ -318,17 +176,19 @@ def show(self, attributes='snr', mode='imshow', return_figure=False, short_title
...
             Mode to display images.
         return_figure : bool
             Whether to return the figure.
         short_title : bool
-            Whether to use only attribute names as titles for axis.
+            Whether to use only attribute names as titles for subplots.
         savepath : str, optional
             Path to save the figure. `**` is changed to a field base directory, `*` is changed to field base name.
+        bbox : bool
+            Whether crop horizon by its bounding box or not.
```
HollowPrincess, line 188: Maybe it would be better to make another naming? It makes me a little confused with bbox from the Field.

seismiqb/src/geometry/base.py
```diff
@@ -557,6 +557,11 @@ def shape(self):
         """ Cube 3D shape. Same API, as NumPy. """
         return tuple(self.cube_shape)
 
+    @property
+    def bbox(self):
+        """ Bounding box that define field limits. """
+        return np.array([[0, max] for max in self.shape])
```
HollowPrincess, line 563: @cdtn , @SergeyTsimfer Is it important to have this property? 
I see only two use cases of ```Field.bbox```:  in visualizations and in ```verify``` in the ```load_charisma```. And in these cases ```Field.shape``` looks sufficient.

seismiqb/src/labels/horizon_attributes.py
```diff
@@ -472,7 +473,7 @@ def load_attribute(self, src, location=None, use_cache=True, enlarge=False, **kw
     @transformable
     def get_property(self, src, **_):
         """ Load a desired instance attribute. Decorated to allow additional postprocessing steps. """
-        data = getattr(self, src, None)
+        data = copy(getattr(self, src, None))
         if data is None:
             aliases = list(self.ALIAS_TO_ATTRIBUTE.keys())
             raise ValueError(f'Unknown `src` {src}. Expected a matrix-property or one of {aliases}.')
```
HollowPrincess, line 479: Why copy? Maybe we need a special flag here or/and in ```load_attribute``` for copy control?

seismiqb/src/plotters.py
```diff
@@ -208,25 +208,97 @@ def make_nested_data(data, separate):
...
+        """ Infer number of columns or/and rows for ploting provided number of subplots. """
+        ncols = params.pop('ncols', None)
+        nrows = params.pop('nrows', None)
+
+        ceil_div = lambda a, b: -(-a // b)
+        if ncols is None and nrows is None:
+            ncols = min(default_ncols, n_subplots)
+            nrows = ceil_div(n_subplots, ncols)
+        elif ncols is None:
+            ncols = ceil_div(n_subplots, ncols)
```
HollowPrincess, line 268:
```suggestion
        elif ncols is None:
            ncols = ceil_div(n_subplots, nrows)
```

seismiqb/src/field/visualization.py
cdtn: Because it's not an actual colormap name, but a value indicating the color should be generated.

seismiqb/src/field/visualization.py
cdtn: Code refactored for clarity, see new commits.

seismiqb/src/field/visualization.py
cdtn: Why not?

seismiqb/src/field/visualization.py
cdtn: Both keys must always be present in params.

seismiqb/src/field/visualization.py
cdtn: image

seismiqb/src/geometry/base.py
cdtn: Either way, `VisualizationMixin` needs the `geometry`'s bounding box to display its attributes correctly with respect to data from other sources. The question is only about where the calculation will be performed. From the point of code uniformity view I prefer to place it in geometry.

seismiqb/src/labels/horizon_attributes.py
cdtn: Since `get_property` is decorated with `transformable` it will modify any cached matrices in place, which we'd all very much like to avoid.

seismiqb/src/field/visualization.py
HollowPrincess: And other parameters which are in this file?

seismiqb/src/field/visualization.py
HollowPrincess: Sorry, in ```Horizon```

seismiqb/src/geometry/base.py
HollowPrincess: Sorry, I didn't pay enough attention to the path. It is used only once in ```VisualizationMixin```. 
Maybe you are right, but I don't see that this is reusable.

seismiqb/src/utils/classes.py
```diff
@@ -728,7 +728,6 @@ def flat(self):
         return self.flatten()
 
 
-
```
SergeyTsimfer, line 731: R.I.P. `NamedArray`

seismiqb/src/field/visualization.py
```diff
@@ -170,139 +160,8 @@ def show_points(self, src='labels', **kwargs):
...
-            output, label = method(**load)
-            output = output.squeeze()
-            output = postprocess(output)
-
-        attribute_dict['output'] = output
-        attribute_dict['label_instance'] = UniterableProxy(label)
-        return attribute_dict
-
-    @staticmethod
-    def _show_add_plot_defaults(attribute_dict):
```
SergeyTsimfer, line 261: Why split this method into multiple?

seismiqb/src/field/visualization.py
cdtn: Well, that's intended. The `bbox` parameter of the `show` method toggles "crop by boundaries" behavior, which relies on mentioned `bbox` attribute of `geometry` and `horizon`.

seismiqb/src/plotters.py
```diff
@@ -208,25 +219,126 @@ def make_nested_data(data, separate):
...
+            nrows = ceil_div(n_subplots, ncols)
+
+        return ncols, nrows
+
+    @classmethod
+    def make_or_parse_axes(cls, data, mode, shapes, all_params):
+        """ Create figure and axes if needed, else use provided. """
+        axes = all_params.pop('axes', None)
+        axes = all_params.pop('axis', axes)
         axes = all_params.pop('ax', axes)
```
HollowPrincess, line 318: Maybe:
```
axes = list(map(all_params.get,  ['axes', 'axis', 'ax']))
axes = sorted(axes , key=lambda x: (x is None, x))[0]
```

or

```
axes = list(map(lambda x: all_params.pop(x, None), ['axes', 'axis', 'ax']))
axes = sorted(axes, key=lambda x: (x is None, x))[0]
```
?

SergeyTsimfer (approved):

PR #148, opened 2021-11-03
Make color lists inexhaustible via loop indexing

cdtn:


seismiqb/src/utils/classes.py
```diff
@@ -553,6 +553,35 @@ def from_grid(cls, grid, aggregation='max', dtype=np.float32, transform=None, pa
                    shape=grid.shape, origin=grid.origin, orientation=grid.orientation, **kwargs)
 
 
+class LoopedList(list):
```
SergeyTsimfer, line 556: `from itertools import cycle`?

SergeyTsimfer (approved):

PR #152, opened 2021-11-10
Fix fill_to_num  unexpected behaviour when target matrix is bool

cdtn:


cdtn:
Also add `fill_value` arg to a few `CropBatch` methods

SergeyTsimfer (approved):

PR #156, opened 2021-11-26
Cache mixin

HollowPrincess:


seismiqb/src/cache.py
```diff
@@ -0,0 +1,110 @@
+""" A cache mixin. """
```
SergeyTsimfer, line 1: Move it to `utils/cache.py`, and move the `lru_cache` decorator there as well

SergeyTsimfer: And re-check the docstrings of both this mixin and `lru_cache` class itself

SergeyTsimfer (approved):

SergeyTsimfer (approved):

PR #158, opened 2021-12-07
Add grid extension and filter_outliers in quality_grid

HollowPrincess:
Add extension for grid lines and filter outliers on the quality map for better quality grid views.

Old view:
image

New view (```extension = 'cell'``` and ```filter_outliers=100```):
image

```filter_outliers``` filter regions with small connectivity objects from the quality map.
image

The ```extension``` parameter extends lines from points on the quality map.

Possible modes: 
 - 'full' for extension to a field borders.
 - 'cell' for extension to a grid cell borders.
 - False - no extension.
 - int number extends to the neighbouring points in range +-number//2.

### DIFF
--- a/seismiqb/src/functional.py
+++ b/seismiqb/src/functional.py
@@ -314,10 +314,47 @@ def digitize(matrix, quantiles):
     digitized[np.isnan(matrix)] = np.nan
     return digitized
 
-def gridify(matrix, frequencies, iline=True, xline=True, full_lines=True, elongation=0, filter_outliers=0):
-    """ Convert digitized map into grid with various frequencies corresponding to different bins. """
-    elongation //= 2
+def elongate_grid(grid, idx_1, idx_2, max_idx_2, transposed, elongation, max_frequency, cut_lines_by_grid):
+    """ Elongate lines on grid on elongation value and cut them if needed.
 
+    Parameters:
+    ----------
+    grid : ndarray
+        A 2D array with grid matrix.
+    idx_1, idx_2 : ndarray
+        Arrays of quality grid points indices by axes.
+    max_idx_2 : int
+        Max possible coordinate of grid point for second array.
+    transposed : bool
+        Whether the idx_1, idx_2 are transposed in relation to the coordinate axes.
+    elongation : int
+        Amount of traces to elongate near the trace on quality map.
+    max_frequency : int
+        Grid frequency for the simplest level of hardness in `quality_map`.
+    cut_lines_by_grid : bool
+        Whether to cut lines by sparse grid.
+    """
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: seismiqb/src/functional.py [hunk 1]
```diff
@@ -336,14 +342,26 @@ def gridify(matrix, frequencies, iline=True, xline=True, full_lines=True):
             if full_lines:
                 grid[idx_1[mask], :] = 1
             else:
-                grid[idx_1[mask], idx_2[mask]] = 1
+                # repeat indices by 1 axis 2*elongation+1 times for idxs consistency
+                idx_1_indices = np.transpose(np.repeat(idx_1[mask], 2*elongation+1).reshape(len(idx_1[mask]), -1))
```
SergeyTsimfer, line 346: I would rather do the same thing, as in `Horizon.add_to_mask`:
```
for shift in range(-elongation, elongation):
    indices_2 = np.clip(idx_2[mask] + shift, 0, matrix.shape[1]-1)
    grid[idx_1[mask], indices_2] = 1
```

seismiqb/src/functional.py
```diff
@@ -313,9 +314,14 @@ def digitize(matrix, quantiles):
     digitized[np.isnan(matrix)] = np.nan
     return digitized
 
-
-def gridify(matrix, frequencies, iline=True, xline=True, full_lines=True):
+def gridify(matrix, frequencies, iline=True, xline=True, full_lines=True, elongation=0, filter_outliers=0):
     """ Convert digitized map into grid with various frequencies corresponding to different bins. """
+    elongation //= 2
+
+    # Preprocess a matrix: drop small complex regions from a quality map to make grid thinned
```
SergeyTsimfer, line 321: What do you think about removing small connected regions? i.e.
```
labeled = label(matrix > 0)
objects = find_objects(labeled)

for object_slice in objects:
   if the number of points in the object is less than threshold, remove
```

The same as in `Horizon.from_mask`

seismiqb/src/functional.py
```diff
@@ -313,9 +315,62 @@ def digitize(matrix, quantiles):
...
+        labeled = label(matrix > 0)
+        objects = find_objects(labeled)
+
+        # Get and vanish too small connectivity regions
+        for object_slice in objects:
+            obj = matrix[object_slice]
+            obj_points = np.sum(obj)
+
+            if obj_points < filter_outliers:
+                matrix[object_slice] = 0
```
SergeyTsimfer, line 371: You should zero-out the individual points in current object, not the entire slice

SergeyTsimfer (approved):

PR #161, opened 2021-12-15
Add filtering crop left or right part

HollowPrincess:


seismiqb/src/crop_batch.py
```diff
@@ -500,15 +500,66 @@ def filter_out(self, ix, src=None, dst=None, expr=None, low=None, high=None, len
...
+        filter_left : bool
+            Whether to filter out the left part of the crop.
+        p : float
+            Probability of applying the transform. Default is 1.
+        """
+        if not (src and dst):
+            raise ValueError('Src and dst must be provided')
+
+        mask = self.get(ix, src)
+        coords = np.where(mask > 0)
```
SergeyTsimfer, line 543: It would be more readable (and also faster) to define slice (either a `:length` or `-length:`) and use it to zero-out the copy of a mask: `mask.copy()[:, slice, :]`

seismiqb/src/crop_batch.py
```diff
@@ -500,21 +500,66 @@ def filter_out(self, ix, src=None, dst=None, expr=None, low=None, high=None, len
...
+
+        if np.random.binomial(1, 1 - p) or len(coords[0]) == 0:
+            # Nothing to apply
+            new_mask = mask
+        else:
+            new_mask = mask.copy()
+
+            # Get the amount of crop lines and kept them on the chosen crop part
+            max_len = mask.shape[0]
+            length = int(np.round(max_len * length_ratio))
```
HollowPrincess, line 553:
```suggestion
            length = int(round(max_len * length_ratio))
```

SergeyTsimfer (approved):

seismiqb/src/crop_batch.py
```diff
@@ -500,21 +500,65 @@ def filter_out(self, ix, src=None, dst=None, expr=None, low=None, high=None, len
...
+
             new_mask[coords[:, 0], coords[:, 1], coords[:, 2]] = mask[coords[:, 0],
                                                                       coords[:, 1],
                                                                       coords[:, 2]]
         return new_mask
 
 
+    @action
+    @inbatch_parallel(init='_init_component', post='_assemble', target='for')
+    def filter_tails(self, ix, src, dst, length_ratio, filter_left, p=1.0):
```
AlexeyKozhevin, line 523: Rewrite it with `apply_parallel` instead of `action` and `inbatch_parallel`.

You don't need to define `p` in your actions with `apply_parallel` decorator: `p` as the probability is available for all actions in batchflow.

HollowPrincess: Thanks)

PR #164, opened 2021-12-29
Extension improvements

HollowPrincess:


seismiqb/src/crop_batch.py
```diff
@@ -500,21 +500,49 @@ def filter_out(self, ix, src=None, dst=None, expr=None, low=None, high=None, len
...
+            Whether to filter out the left part of the crop.
+        """
+        new_mask = np.copy(crop)
+
+        # Get the amount of crop lines and kept them on the chosen crop part
+        max_len = new_mask.shape[0]
+        length = round(max_len * length_ratio)
+
+        if filter_left:
+            new_mask[:-length, :] = 0
```
AlexeyKozhevin, line 539: Let's see an example when `new_mask.shape[0]  == 20`,  `length_ratio == 0.01`, `filter_left == True`.

Then `length` will be 0 and the whole mask will be 0.

seismiqb/src/samplers.py
```diff
@@ -1317,6 +1397,12 @@ def _make_locations(self):
         locations[:, 2:9] = buffer
         self.locations = locations
 
+        if update_coverage_matrix:
+            self.uncovered_best = coverage_matrix.size - coverage_matrix.sum()
+        else:
+            # In the perspective mode we choosed the best variant
+            self.uncovered_best = self.uncovered_after
```
HollowPrincess, line 1404: rethink ` self.uncovered_best`

seismiqb/src/samplers.py
```diff
@@ -1220,6 +1692,26 @@ def __init__(self, horizon, crop_shape, stride=16, batch_size=64, top=1, thresho
...
+            'perspective': possible_directions
+        }
+
+        if self.mode in direction_modes:
+            self.directions = direction_modes[self.mode]
+        elif self.mode in possible_directions:
+            self.directions = [self.mode]
+        else:
+            # Old behavior equals 'various' mode, so we use this mode on default
+            warnings.warn('Provided wrong `mode` argument, it is automatically changed to `various`.')
```
SergeyTsimfer, line 1711: Don't do that: if something looks wrong, fail with an exception

seismiqb/src/samplers.py
```diff
@@ -1220,6 +1692,26 @@ def __init__(self, horizon, crop_shape, stride=16, batch_size=64, top=1, thresho
 
         self.uncovered_before = None
 
+        possible_directions = ['up', 'down', 'left', 'right']
```
SergeyTsimfer, line 1695: `allowed_directions`

seismiqb/src/samplers.py
```diff
@@ -1220,6 +1692,26 @@ def __init__(self, horizon, crop_shape, stride=16, batch_size=64, top=1, thresho
...
+        direction_modes = {
```
SergeyTsimfer, line 1696: You can use fewer code lines to get the same resulting `self.directions`

seismiqb/src/samplers.py
```diff
@@ -1256,57 +1748,102 @@ def _make_locations(self):
...
-        # Crops with fixed CROSSLINE, moving INLINE: [-shape + stride:+stride]
-        buffer[3, :, [1, 4]] -= (crop_shape[1] - self.stride)
-        np.clip(buffer[3, :, 1], 0, self.field.shape[0] - crop_shape[1], out=buffer[3, :, 1])
-        np.clip(buffer[3, :, 4], 0, self.field.shape[0] - crop_shape[1], out=buffer[3, :, 4])
-        buffer[3, :,  [4, 5, 6]] += crop_shape_t.reshape(-1, 1)
-        buffer[3, :, 0] = 1
-
-        if self.randomize:
-            buffer = buffer[np.random.permutation(4)]
+        directions_num = len(self.directions)
```
SergeyTsimfer, line 1751: n_directions?

seismiqb/src/samplers.py
```diff
@@ -1256,57 +1748,102 @@ def _make_locations(self):
...
-        shifts = np.repeat(np.arange(0, len(buffer), 4, dtype=np.int32), self.top)
-        indices = argsort + shifts
+        else:
+            indices = slice(None)
 
-        # Keep only top locations; remove too locations with small potential
+        # Keep only top locations; remove locations with too small potential if needed
         potential = potential[indices]
         buffer = buffer[indices, :]
+        self.choosed_crops = buffer.shape[0]
```
SergeyTsimfer, line 1842: There are no crops, only locations
`n_possible_locations`, `n_top_locations`, `n_selected_locations`

seismiqb/src/samplers.py
```diff
@@ -1256,57 +1748,102 @@ def _make_locations(self):
...
+        self.reviewed_crops = buffer.shape[0]
 
         # Compute potential addition for each location
-        potential = compute_potential(buffer, coverage_matrix, crop_shape)
-        self.uncovered_best = coverage_matrix.size - coverage_matrix.sum()
+        # for 'perspective' and 'independent_various' modes potential calculated independently
+        potential = compute_potential(locations=buffer, coverage_matrix=coverage_matrix,
+                                      shape=crop_shape, update_coverage_matrix=update_coverage_matrix)
+
+        if self.mode.find('various') != -1:
```
SergeyTsimfer, line 1818: does that mean `self.mode == 'various'`?

seismiqb/src/samplers.py
```diff
@@ -1256,57 +1748,102 @@ def _make_locations(self):
...
+
+        elif self.mode == 'perspective':
+            # Get indices of locations corresponding to the most perspective direction
+            # The most perspective direction is a direction with maximal potentially added traces
+            positive_potential = potential.copy()
+            positive_potential[positive_potential < 0] = 0
 
-        # Get argsort for each group of four
-        argsort = potential.reshape(-1, 4).argsort(axis=-1)[:, -self.top:].reshape(-1)
+            perspective_direction_idx = np.argmax(positive_potential.reshape(-1, directions_num).sum(axis=0))
```
SergeyTsimfer, line 1833: Hard to grasp whether this is correct; in any case, that is the same as the previous branch with `self.top=1`

seismiqb/src/samplers.py
SergeyTsimfer, line 1842: chosen

seismiqb/src/samplers.py
```diff
@@ -1256,57 +1748,102 @@ def _make_locations(self):
...
+            argsort = potential.reshape(-1, directions_num).argsort(axis=-1)[:, -self.top:].reshape(-1)
+
+            # Shift argsorts to original indices
+            shifts = np.repeat(np.arange(0, len(buffer), directions_num, dtype=np.int32), self.top)
+            indices = argsort + shifts
+
+        elif self.mode == 'perspective':
+            # Get indices of locations corresponding to the most perspective direction
+            # The most perspective direction is a direction with maximal potentially added traces
+            positive_potential = potential.copy()
```
SergeyTsimfer, line 1830: locations with negative potential will be filtered anyway, right?

SergeyTsimfer (commented):
Move the `weighted_sampler` logic to a separate PR

seismiqb/src/samplers.py
HollowPrincess: No, it means `self.mode in ['various', 'independent_various']`. Fixed it.

seismiqb/src/samplers.py
HollowPrincess: I think it is not: this branch get one direction for all points, previous branch get one direction for each point.
Example:
python draft

seismiqb/src/samplers.py
HollowPrincess: Yes, but in this case, I need a sum for each allowed direction (for finding the best one) and -1 make it worse.

seismiqb/src/samplers.py
```diff
@@ -1206,12 +1206,26 @@ class ExtensionGrid(BaseGrid):
         Whether to randomize the loop for computing the potential of each location.
     batch_size : int
         Number of batches to generate on demand.
+    mode : str
```
SergeyTsimfer, line 1209: Numpy docstyle

seismiqb/src/samplers.py
```diff
@@ -1206,12 +1206,26 @@ class ExtensionGrid(BaseGrid):
...
+        Mode for directions of crops locations to generate.
+        Possible options are: 'up', 'down', 'left', 'right', 'vertical', 'horizontal',
```
SergeyTsimfer, line 1211: Can we come up with better naming? It is really hard to figure out what is going on even with the documentation

PR #168, opened 2022-02-01
Filter masks with discontinuities

HollowPrincess:


seismiqb/src/crop_batch.py
```diff
@@ -545,6 +545,47 @@ def filter_sides(self, crop, length_ratio, side):
...
+        # Get mask and shifted mask for neighboring traces comparison
+        cutted_mask = mask[:, :-1]
+        shifted_mask = mask[:, 1:]
+
+        # Get horizons coordinates on the mask
+        _, x_lines_1, depths_1 = np.where(cutted_mask > 0)
+        _, x_lines_2, depths_2 = np.where(shifted_mask > 0)
+
+        for x in np.unique(x_lines_1):
+            if x in x_lines_2:
```
SergeyTsimfer, line 569: Sounds like you want to do a `groupby` operation on `x`, getting `min`/`max` values for each trace, and then use those for filtering. Vectorized operations should be a lot faster..?

Just in case: we have `njitted` version of `groupby` for np arrays.

SergeyTsimfer (changes requested):
A gentle ping

SergeyTsimfer (approved):

PR #174, opened 2022-02-25
Improve `compute_support` for horizon carcass case

HollowPrincess:


seismiqb/src/metrics.py
```diff
@@ -260,6 +260,22 @@ def compute_support(self, function, data, bad_traces, supports, safe_strip=0,
                 bad_traces_ = bad_traces
 
             valid_traces = xp.where(bad_traces_ == 0)
+
+            if hasattr(self, 'horizon'):
+                if self.horizon.is_carcass:
```
SergeyTsimfer, line 265: This should be one line

seismiqb/src/metrics.py
```diff
@@ -276,14 +292,17 @@ def compute_support(self, function, data, bad_traces, supports, safe_strip=0,
         # Compute metric
         pbar = Notifier(pbar, total=len(support_traces))
         accumulator = Accumulator(agg=agg, amortize=amortize, axis=axis, total=len(support_traces))
+
         for i, _ in enumerate(support_traces):
-            computed = function(data_n, support_traces[i], data_stds, support_stds[i])
-            computed[bad_traces == 1] = xp.nan
+            computed = function(data_n[bad_traces != 1], support_traces[i],
```
SergeyTsimfer, line 297: slicing for `data_n` array should be done only once before the loop

PR #178, opened 2022-03-17
Update codebase to work with latest `plot`

cdtn:


SergeyTsimfer (commented):
This should be merged after:
- next batchflow release
- moving away from `batchflow` as a submodule

PR #182, opened 2022-03-29
Inspect the `Horizon` class

HollowPrincess:


seismiqb/labels/horizon_processing.py
```diff
@@ -0,0 +1,283 @@
...
+from scipy.ndimage.morphology import binary_fill_holes, binary_dilation
+
+from ..functional import smooth_out
+from ..utils import make_bezier_figure
+
+class ProcessingMixin:
+    """Methods for horizon processing.
+
+    These methods transform horizon structure and storages: matrix and/or points.
+    The main classes of these methods are:
```
SergeyTsimfer, line 15: 🤔

seismiqb/functional.py
```diff
@@ -163,72 +164,178 @@ def make_gaussian_kernel(kernel_size=3, sigma=1.):
...
-                     fill_value=None, preserve=True, margin=np.inf, **_):
+def process_missings(function):
+    """ Decorator which apply a special treatment for the missing points in a `matrix`
+    which marked with either `fill_value` or `np.nan`. """
+    @wraps(function)
+    def wrapper(matrix, fill_value=None, **kwargs):
+        # Parse arguments
+        preserve_missings = kwargs.get('preserve_missings', False)
+
+        if function.__name__ == 'interpolate':
```
SergeyTsimfer, line 175: Ugly
Could be moved to decorator parameters / method defaults. In the latter case, you would need to use `inspect` to get the method signature and merge `method.signature.defaults` with `kwargs`

seismiqb/functional.py
```diff
@@ -163,72 +164,178 @@ def make_gaussian_kernel(kernel_size=3, sigma=1.):
...
-
-def special_convolve(matrix, mode='convolve', kernel_size=3, kernel=None, iters=1,
-                     fill_value=None, preserve=True, margin=np.inf, **_):
+def process_missings(function):
+    """ Decorator which apply a special treatment for the missing points in a `matrix`
+    which marked with either `fill_value` or `np.nan`. """
+    @wraps(function)
+    def wrapper(matrix, fill_value=None, **kwargs):
+        # Parse arguments
+        preserve_missings = kwargs.get('preserve_missings', False)
```
SergeyTsimfer, line 173: Move to the signature explicitly

seismiqb/functional.py
```diff
@@ -238,71 +345,103 @@ def _convolve(src, kernel, preserve, margin):
...
+        for xline in range(k, x_range - k):
+            central = src[iline, xline]
+
+            if not isnan(central):
+                continue # We interpolate values only to nan points
+
+            # Get neighbors and check whether we can interpolate them
+            element = src[iline-k:iline+k+1, xline-k:xline+k+1]
+
+            notnan_neighbors = kernel.size - np.nansum(np.isnan(element))
```
SergeyTsimfer, line 384: `np.isnan(element).sum` is enough -- that is a `bool`-dtyped matrix.


Almost sure that a loop like:
```python
from math import isnan

present_neightbours = kernel.size
for e in element:
    if isnan(e):
        present_neightbours -= 1
```
would be faster, although might not be worth it

seismiqb/labels/horizon_processing.py
```diff
@@ -0,0 +1,356 @@
...
+    All of these methods affect horizon structure, so be careful with them: their result is a new horizon surface.
+    """
+    # Filtering methods
+    def filter_points(self, filtering_matrix=None, margin=0, **_):
+        """ Remove points that correspond to 1's in `filtering_matrix` from points storage. """
+        if filtering_matrix is None:
+            filtering_matrix = self.field.zero_traces
+        if margin > 0:
+            filtering_matrix = binary_dilation(filtering_matrix, structure=np.ones((margin, margin)))
+            filtering_matrix[:margin, :] = 1
```
SergeyTsimfer, line 26: `filtering_matrix[-margin:margin, :] = 1` ?

seismiqb/functional.py
```diff
@@ -163,72 +164,178 @@ def make_gaussian_kernel(kernel_size=3, sigma=1.):
     gaussian_kernel = (kernel / np.sum(kernel).astype(np.float32))
     return gaussian_kernel
 
-
-def special_convolve(matrix, mode='convolve', kernel_size=3, kernel=None, iters=1,
-                     fill_value=None, preserve=True, margin=np.inf, **_):
+def process_missings(function):
```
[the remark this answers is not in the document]
HollowPrincess: I think this is more about matrices, than about horizons.

seismiqb/labels/horizon_processing.py
HollowPrincess: It is not work

seismiqb/functional.py
```diff
@@ -238,71 +345,103 @@ def _convolve(src, kernel, preserve, margin):
...
+
+            # Get neighbors and check whether we can interpolate them
+            element = src[iline-k:iline+k+1, xline-k:xline+k+1]
+
+            notnan_neighbors = kernel.size - np.nansum(np.isnan(element))
+            if notnan_neighbors < min_neighbors:
+                continue
+
+            if margin is not None: # Is the difference between values in the window too high
+                nanmax, nanmin = np.nanmax(element), np.nanmin(element)
```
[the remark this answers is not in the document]
HollowPrincess: Is slower (`_inerpolate` - implementation in PR, `new_interpolate` - proposed, matrix shape = (1000, 1000), kernel size =3)

image

seismiqb/functional.py
HollowPrincess: Is not faster (_inerpolate - implementation in PR, new_interpolate - proposed, matrix shape (1000, 1000), kernel size 3)
image

seismiqb/labels/horizon_processing.py
```diff
@@ -0,0 +1,400 @@
...
+        uniques, counts = np.unique(self.points[:, 0], return_counts=True)
+        mask_i = np.isin(self.points[:, 0], uniques[counts > threshold][::factor[0]])
+
+        uniques, counts = np.unique(self.points[:, 1], return_counts=True)
+        mask_x = np.isin(self.points[:, 1], uniques[counts > threshold][::factor[1]])
+
+        self.points = self.points[mask_i + mask_x]
+        self.reset_storage('matrix')
+
+    def smooth_out(self, kernel=None, kernel_size=3, iters=1, preserve_missings=True, distance_threshold=5,
```
SergeyTsimfer, line 129: Currently, this function works in a sense that it fulfills its intended purpose; yet, it raises a lot of questions about its implementations / usage / interface / beauty:
- starting from signature. Why this order of parameters? `kernel, kernel_size, sigma` should be one after the other, maybe even in `kernel_size, sigma, kernel` order.
- do we even need `preserve_missings` parameter? One can use `interpolate -> smooth_out` to emulate the same, while being more explicit.
- do we even need `kernel` parameter? Probably yes.
- `distance_threshold`. what is the `distance`? be more specific: ~`depth_threshold`
- do we need `**_`? Probably not.
- the implementation definitely calls for moving the body of `smooth_out` to here, especially with the fact that this `smooth_out` function does not do anything but makes a kernel, and re-direct to another (`convolve`) function, which actually once again re-makes the kernel, if not provided, and one more time re-directs to a `_convolve`. The number of `re-directs` in the previous sentence should not be 3.

seismiqb/labels/horizon_processing.py
```diff
@@ -0,0 +1,400 @@
...
+                              distance_threshold=distance_threshold, sigma=sigma)
+
+        smoothed = np.rint(smoothed).astype(np.int32)
+        smoothed[self.field.zero_traces[self.i_min:self.i_max + 1,
+                                        self.x_min:self.x_max + 1] == 1] = self.FILL_VALUE
+
+        self.matrix = smoothed
+        self.reset_storage('points')
+
+    def edge_preserving_smooth_out(self, kernel_size=3, sigma_spatial=0.8, iters=1,
```
SergeyTsimfer, line 166: If one tracks down the sequence of re-directs of this function to the core, it becomes clear, that the only difference to the previous one is that it uses `_bilateral_filter` instead of `_convolve`. `smooth_out(..., mode='bilateral/convolve`) looks better and saves a lot of space.

SergeyTsimfer (changes requested):
Each of the `smooth_out, smooth_out_bilateral, interpolate, median_filter` is obscured by a number of unnecessary levels of function calls and decorators and being split across multiple files.

Each of them is the same crystal-clear sequence of actions:
```python
# Parse parameters, e.g. prepare kernel if it is not passed explicitly

# Pre-process self.matrix: change fill_value, apply padding, save the mask of present/absent values (in a separate named variable!)

# Apply some function multiple times

# Post-process the output: strip padding, return fill_values, set correct type

# Set the matrix to a horizon in-place (and reset points storage) or return it explicitly
```

Some parts can be moved to a decorator or become separate methods; nevertheless, it should be clear from reading the code in one place.

Please, think out and write down a cohesive plan before touching actual code. Current working version of functions is really just an MVP of what it should be: a coherent piece of clearly written code.

seismiqb/labels/horizon_attributes.py
```diff
@@ -643,71 +644,167 @@ def get_zerocrossings(self, side, window=15):
...
             spikes = np.nan_to_num(spikes)
             spikes = binary_dilation(spikes, iterations=dilation).astype(np.float32)
             spikes[self.field.zero_traces == 1] = np.nan
+
         return spikes
+
+# Helper functions
+@njit(parallel=True)
+def _spikes_along_axis(matrix, axis=0, spike_max_width=5, spike_min_height=3, close_depths_threshold=2):
+    """ Get spikes mask along axis. """
```
SergeyTsimfer, line 748: having a `check_line` function and running it twice: `check_line(line) and check_line(line[::-1]` would make things a lot easier

seismiqb/labels/horizon_attributes.py
```diff
@@ -643,71 +644,167 @@ def get_zerocrossings(self, side, window=15):
...
+            grad = binary_dilation(grad, iterations=dilation).astype(np.float32)
+            grad[self.field.zero_traces == 1] = np.nan
+        return grad
 
     @lru_cache(maxsize=1, apply_by_default=False, copy_on_return=True)
     @transformable
-    def get_spikes_map(self, spikes_mode='median', threshold=1., dilation=5,
-                       kernel_size=11, kernel=None, margin=0, iters=2, **_):
-        """ Locate spikes on a horizon.
+    def get_spikes_mask(self, spike_max_width=7, spike_min_height=5, close_depths_threshold=2, dilation=0):
```
SergeyTsimfer, line 707: Please, don't use the `height` word when it regards `depth`

`min/max` are misleading in a way that it is hard to understand, whether `min_height` is minimum allowed depth of a spike or is it the minimum detected depth for a spike to be detected


`width_threshold`, maybe?

`depth1_threshold` / `depth2_threshold`? Those two names force you to look into parameter documentation and don't rely on naming being intuitively clear, which it is not

SergeyTsimfer (changes requested):

HollowPrincess:
Fix geometry tests parametrization

seismiqb/metrics.py
```diff
@@ -862,8 +857,8 @@ def probs(self):
 
 
     def quality_map(self, quantiles, metric_names=None, computed_metrics=None,
```
SergeyTsimfer, line 859: Revert the changes to this file, please. Just make a stub instead of `smooth_out`; we will deal with this right before merging this PR

seismiqb/utils/decorators.py
```diff
@@ -27,6 +30,9 @@ def transformable(method):
         If `min-max` or True, then use min-max scaling.
         If `mean-std`, then use mean-std scaling.
         If False, don't scale matrix.
+    dilation_iterations : int
```
SergeyTsimfer, line 33: `dilation`; default should be `None` or `False`

seismiqb/utils/decorators.py
```diff
@@ -35,26 +41,38 @@ def transformable(method):
...
             channels = to_list(channels)
             result = result[:, :, channels]
+
         if normalize and hasattr(instance, 'matrix_normalize'):
             result = instance.matrix_normalize(result, normalize)
+
+        if dilation_iterations:
+            result = np.nan_to_num(result)
+            result = binary_dilation(result, iterations=dilation_iterations).astype(np.float32)
+            result[instance.field.zero_traces == 1] = np.nan
```
SergeyTsimfer, line 65: All of the other operations in this decorator are applied through `instance` methods, and there is a reason for it.

This forces the `dtype` to be `np.float32`, even if `dtype` argument was passed as any other.

seismiqb/labels/horizon_attributes.py
```diff
@@ -640,74 +675,275 @@ def get_zerocrossings(self, side, window=15):
...
         grad[self.field.zero_traces == 1] = np.nan
         return grad
 
-
     @lru_cache(maxsize=1, apply_by_default=False, copy_on_return=True)
     @transformable
-    def get_spikes_map(self, spikes_mode='median', threshold=1., dilation=5,
-                       kernel_size=11, kernel=None, margin=0, iters=2, **_):
-        """ Locate spikes on a horizon.
+    def get_spikes_mask(self, spike_spatial_maxsize=7, spike_depth_minsize=5, close_depths_threshold=2,
```
SergeyTsimfer, line 751: Why previous methods are `*_map`s, and this one is `_mask`?

seismiqb/labels/horizon_attributes.py
```diff
@@ -640,74 +675,275 @@ def get_zerocrossings(self, side, window=15):
...
+                        dilation_iterations=0):
+        """ Get spikes mask for the horizon.
+
+        We suppose that spikes are huge depth changes with invert depth changes in a fixed size window.
```
SergeyTsimfer, line 755: This is very hard to read, and even harder to understand

seismiqb/labels/horizon_attributes.py
```diff
@@ -640,74 +675,275 @@ def get_zerocrossings(self, side, window=15):
...
         ----------
-        mode : str
-            If 'gradient', then use gradient map to locate spikes.
-            If 'median', then use median diffs to locate spikes.
-        threshold : number
-            Threshold to consider a difference to be a spike.
-        dilation : int
+        spike_spatial_maxsize : int
+            Maximum possible spike size along the iline or xline axes.
+        spike_depth_minsize : int
```
SergeyTsimfer, line 762: `spike_min_depth`?

seismiqb/labels/horizon_processing.py
```diff
@@ -0,0 +1,494 @@
...
+        Returns
+        -------
+        :class:`~.Horizon`
+            Processed horizon instance. A new instance if `inplace` is False, `self` otherwise.
+        """
+        if filtering_matrix is None:
+            filtering_matrix = self.field.zero_traces
+
+        if margin > 0:
+            filtering_matrix = binary_dilation(filtering_matrix, structure=np.ones((margin, margin)))
```
SergeyTsimfer, line 49: That is wrong: for example, if `margin==1`, this does nothing. You should pass `margin` as `iters` parameter

seismiqb/labels/horizon_processing.py
```diff
@@ -0,0 +1,494 @@
...
+            Can be used for separate smoothening on sides of discontinuity.
+        inplace : bool
+            Whether to apply operation inplace or return a new Horizon object.
+
+        Returns
+        -------
+        :class:`~.Horizon`
+            Processed horizon instance. A new instance if `inplace` is False, `self` otherwise.
+        """
+        result = self.matrix_smooth_out(matrix=self.matrix, mode=mode, iters=iters,
```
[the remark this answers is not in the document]
SergeyTsimfer: And also move `njit` functions to here

seismiqb/labels/horizon_processing.py
```diff
@@ -0,0 +1,494 @@
...
+        kernel_size : int
+            If the kernel is not provided, shape of the square gaussian kernel.
+        sigma : number
+            Standard deviation (spread or “width”) for gaussian kernel.
+            The lower, the more weight is put into the point itself.
+        kernel : ndarray or None
+            Kernel to apply to missing points.
+        min_filled_neighbors: int
+            Minimal amount of non-missing neighboring points in a window to interpolate a central point.
+        depths_variance_threshold : number
```
SergeyTsimfer, line 240: `max_depth_ptp`

SergeyTsimfer (changes requested):
Rethink naming of thresholds, please

seismiqb/labels/horizon_attributes.py
SergeyTsimfer: or, better yet, `min_spike_depth` and `max_spike_length`

seismiqb/labels/horizon.py
```diff
@@ -1160,10 +846,12 @@ def dump_float(self, path, transform=None, kernel_size=7, sigma=2., margin=5):
...
+            then the point is ignored in smoothening.
+            Can be used for separate smoothening on sides of discontinuity.
         """
-        matrix = self.matrix_smooth_out(matrix=self.full_matrix, kernel_size=kernel_size, sigma=sigma, margin=margin)
-        points = self.matrix_to_points(matrix)
-        self.dump_charisma(data=points, path=path, format='points', name=self.name, transform=transform)
+        smoothed = self.smooth_out(mode='convolve', kernel_size=kernel_size, sigma_spatial=sigma,
+                                   max_depth_difference=max_depth_difference, inplace=False)
+
+        self.dump_charisma(data=smoothed.points, path=path, format='points', name=self.name, transform=transform)
```
SergeyTsimfer, line 857: The idea of this method is to save floating-points number, produced by smoothing out integer values, on disk.

This implementation converts values back to integer and is effectively equivalent to separate method calls `smooth_out(...) -> dump(...)`

seismiqb/labels/horizon_processing.py
```diff
@@ -0,0 +1,645 @@
...
+        if erosion_rate > 0:
+            filtering_matrix = binary_dilation(filtering_matrix, structure, iterations=erosion_rate)
+
+        filtering_matrix = filtering_matrix == 0
+
+        return self.filter(filtering_matrix, inplace=inplace)
+
+
+    # Horizon surface transformations
+    def thin_out(self, factor=1, threshold=256, inplace=True):
```
SergeyTsimfer, line 122: Reposition methods, so that `thin_out` is near `make_carcass/make_holes`

seismiqb/labels/horizon_processing.py
```diff
@@ -0,0 +1,645 @@
...
+
+        if inplace:
+            self.points = points
+            self.reset_storage('matrix')
+            return self
+
+        name = 'filtered_' + self.name if self.name is not None else None
+        return type(self)(storage=points, field=self.field, name=name)
+
+    def filter_spikes(self, spike_spatial_maxsize=7, spike_depth_minsize=5, close_depths_threshold=2,
```
SergeyTsimfer, line 75: If all of the parameters are default to `spikes` attribute, we can not just call `filter('spikes')`, right?

seismiqb/labels/horizon_processing.py
```diff
@@ -0,0 +1,645 @@
...
+        name = 'thinned_' + self.name if self.name is not None else None
+        return type(self)(storage=points, field=self.field, name=name)
+
+    def smooth_out(self, mode='convolve', iters=1,
+                   kernel_size=3, sigma_spatial=0.8, kernel=None, sigma_range=2.0,
+                   max_depth_difference=5, inplace=True):
+        """ Smooth out the horizon surface.
+
+        Smoothening is applied with a special treatment to absent points:
+            - if `preserve_missings=True`, then missing points remain absent in the resulting surface.
```
SergeyTsimfer, line 167: There is no such parameter

seismiqb/labels/horizon_processing.py
```diff
@@ -0,0 +1,645 @@
...
+            present points nearby.
+
+        This method supports two types of smoothening:
+            - if `mode='convolve'`, then the method uses a convolution with a given or a gaussian kernel.
+            - if `mode='bilateral'`, then the method applies a bilateral filtering with a given or a gaussian kernel.
+        Bilateral filtering is an edge-preserving smoothening, which ignores areas with faults.
+
+        Be careful with `sigma_range` value:
+            - The higher the `sigma_range` value, the more 'bilateral' result looks like a 'convolve' result.
+            - If the `sigma_range` too low, then no smoothening applied.
```
SergeyTsimfer, line 178: That is a comment only to `bilateral`. Indent it

seismiqb/labels/horizon_processing.py
```diff
@@ -0,0 +1,645 @@
...
+
+        Note, this method may change horizon inplace or create a new instance. By default works inplace.
+        In either case it returns a processed horizon instance.
```
SergeyTsimfer, line 181: Returning a `float32` matrix if `not inplace` would make lives (yours in particular) a lot easier.

On the other hand, you can try `Horizon(..., dtype=np.float32)` and see if it works

seismiqb/labels/horizon_processing.py
```diff
@@ -0,0 +1,645 @@
...
+        :class:`~.Horizon`
+            Processed horizon instance. A new instance if `inplace` is False, `self` otherwise.
+        """
+        if 'conv' in mode:
+            smoothening_function, kwargs = _convolve, {}
+        else:
+            smoothening_function, kwargs = _bilateral_filter, {'sigma_range': sigma_range}
+
+        if kernel is None:
+            kernel = make_gaussian_kernel(kernel_size=kernel_size, sigma=sigma_spatial)
```
SergeyTsimfer, line 220: `kernel = kernel if kernel is not None else ...`?

seismiqb/labels/horizon_processing.py
```diff
@@ -0,0 +1,634 @@
...
+        if self.dtype==np.int32 or (dtype == np.int32 and inplace is False):
+            result = np.rint(result).astype(np.int32)
+
+        if inplace:
+            self.matrix = result
+            self.reset_storage('points')
+            return self
+
+        name = 'smoothed_' + self.name if self.name is not None else None
+        return type(self)(storage=result, i_min=self.i_min, x_min=self.x_min, field=self.field, name=name, dtype=dtype)
```
SergeyTsimfer, line 193: Is the dumped horizon actually float?

PR #189, opened 2022-05-12
Add evaluate supports method for a horizons list

HollowPrincess:


seismiqb/labels/horizon.py
```diff
@@ -974,143 +973,13 @@ def find_closest(self, *others):
...
-
-            MatplotlibPlotter.save_and_show(fig=fig, show=show, savepath=savepath)
-            if return_figure:
-                returns['figure'] = fig
-
-        return returns
+    # Alias for horizon comparisons
+    def compare(self, *others, clip_value=5, ignore_zeros=True,
+                printer=print, plot=True, return_figure=False, hist_kwargs=None, **kwargs):
+        """ Alias for `Horizon.compare`. """
```
SergeyTsimfer, line 979: `HorizonMetrics.compare`.
Also, use `wraps` or other ways of keeping the correct docstring in this method

seismiqb/metrics.py
```diff
@@ -620,6 +680,70 @@ def __init__(self, horizons, window=23, offset=0, normalize=False, chunk_size=25
...
+
+            support_coords = xp.asarray([valid_traces[0][indices], valid_traces[1][indices]]).T
+
+        elif isinstance(supports, (tuple, list, np.ndarray)):
+            support_coords = xp.asarray(supports)
+
+        # Evaluate support metric for all horizons
+        metrics = []
+
+        for idx, horizon in enumerate(horizons):
```
SergeyTsimfer, line 726: Make this method a `classmethod` and use `cls(horizon).evaluate(..., supports=support_coords)` instead of manually doing everything.

This way, you also would not need to pass `data` to any of `support_***` method, so revert this as well

seismiqb/metrics.py
```diff
@@ -620,6 +623,53 @@ def __init__(self, horizons, window=23, offset=0, normalize=False, chunk_size=25
...
+    @classmethod
+    def evaluate_support(cls, horizons, metric='support_corrs', bad_traces=None, supports=100, safe_strip=0,
+                         device='cpu', seed=None, **kwargs):
+        """ ..!!.. """
+        xp = cp if (CUPY_AVAILABLE and device == 'gpu') else np
+
+        horizons_bad_traces = []
+
+        # Generate support coordinates
+        if isinstance(supports, int):
```
SergeyTsimfer, line 635: Maybe, make a `compute_supports(supports: int, (tuple, list, np.ndarray), bad_traces, safe_strip)` method, which is called both here and in regular support metric computation?

SergeyTsimfer (approved):

PR #206, opened 2022-08-05
Change masks/predictions coloring in batch plot

cdtn:
Never hide mask zeros on plot if its a first subplot layer.
Never hide prediction values lower than 0.5 if its a first layer.

### DIFF
--- a/seismiqb/batch/visualization_batch.py
+++ b/seismiqb/batch/visualization_batch.py
@@ -211,12 +211,8 @@ class VisualizationMixin:
         # Make titles for individual axis
         title = [', '.join(item) for item in components]
         if 'title' in add_location:
-            if len(components) >= 1:
-                title[0] += '\n' + location_info[0]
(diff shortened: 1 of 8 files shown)


seismiqb/batch/visualization_batch.py
```diff
@@ -1,220 +1,362 @@
...
+            if 'suptitle' in add_location:
+                suptitle += '\n' + '   '.join(location_info)
+            config['suptitle'] = suptitle
+
+        # Make titles for individual axis
         title = [', '.join(item) for item in components]
-        # TODO: Replace with `set_xticklabels` parametrization
-        if augment_title:
+        if 'title' in add_location:
             if len(components) >= 1:
```
SergeyTsimfer, line 214:
```python
for title_, location_info_ in zip(title, location_info):
    title_ += '\n' + location_info_
```
?

PR #224, opened 2022-10-14
Refactor cache

HollowPrincess:
For a big amount of horizons `CacheMixin.get_cached_objects` (in cache reset) call took a long time because of iterating on `horizon.__dir__()` objects.
Cache logic was changed: now all cache stored in the object (in the `cache` attribute).
New implementation controls only methods and properties which use `lru_cache` decorator from the **seismiqb**.

Also, add the `GlobalCache` object for global cache introspection and cleaning.
`GlobalCache` implemented through weak references, so it is properly works with deleted elements.

### DIFF
--- a/seismiqb/utils/cache.py
+++ b/seismiqb/utils/cache.py
@@ -28,22 +28,54 @@ class _GlobalCacheClass:
     @property
     def size(self):
         """ Total cache size. """
-        cache_size = 0
-
-        for instance in self.instances_with_cache:
-            cache_size += instance.cache_size
-
-        return cache_size
+        return self.get_stats(stat='size', level='total')
 
     @property
     def nbytes(self):
         """ Total cache nbytes. """
-        cache_nbytes = 0
+        return self.get_stats(stat='nbytes', level='total')
 
+    def get_size(self, level='total'):
+        """ Get cache size grouped by level. For more read the doc for :meth:`~.get_attr`"""
+        return self.get_stats(stat='size', level=level)
+
+    def get_nbytes(self, level='total'):
+        """ Get cache nbytes grouped by level. For more read the doc for :meth:`~.get_attr`"""
+        return self.get_stats(stat='nbytes', level=level)
+
+    def get_stats(self, stat='size', level='total'):
+        """ Get cache statistics grouped by level.
+
+        Parameters
+        ----------
+        stat : {'size', 'nbytes'}
+            Statistic to get values.
+        level : {'total', 'class', 'instance'}
+            Result groupby level.
+            If 'total', then return a total stat value for all instances.
+            If 'class', then return a dict with stat value for each class.
+            If 'instance', then return a nested dict with stat value for each instance.
+        """
+        # Init result accumulator/container
+        if level == 'total':
+            result = 0
+        elif level == 'class':
+            result = defaultdict(int)
+        else:
+            result = defaultdict(lambda: defaultdict(int))
+
+        # Fill accumulator/container with attribute values from instances with cache
         for instance in self.instances_with_cache:
-            cache_nbytes += instance.cache_nbytes
+            value = getattr(instance, f'cache_{stat}')
 
-        return cache_nbytes
+            if level == 'total':
+                result += value
+            elif level == 'class':
+                result[instance.__class__.__name__] += value
+            else:
+                result[instance.__class__.__name__][f'id_{id(instance)}'] += value
+
+        return result
 
     def get_cache_repr(self, format='dict'):
         """ Create global cache representation.
--- a/seismiqb/utils/cache.py
+++ b/seismiqb/utils/cache.py
@@ -211,11 +211,7 @@ class lru_cache:
             # Init cache and reference on it in the GlobalCache controller
             if not hasattr(instance, 'cache'):
                 # Init cache container in the instance
-                instance.__setattr__('cache', {})
-
-            if self.attrname not in instance.cache:
-                # Init property/method cache in the cache container
-                instance.cache[self.attrname] = OrderedDict()
+                instance.__setattr__('cache', defaultdict(OrderedDict))
 
             GlobalCache.instances_with_cache.add(instance)
 
@@ -307,7 +303,7 @@ class CacheMixin:
             names = (name,) if name is not None else self.cache.keys()
 
             for attrname in names:
-                cached_values = self.cache.get(attrname, {}).values()
+                cached_values = self.cache[attrname].values()
 
                 cache_length_accumulator += len(cached_values)
 
@@ -328,7 +324,7 @@ class CacheMixin:
             names = (name,) if name is not None else self.cache.keys()
 
             for attrname in names:
-                cached_values = self.cache.get(attrname, {}).values()
+                cached_values = self.cache[attrname].values()
 
                 for value in cached_values:
                     if isinstance(value, np.ndarray):
--- a/seismiqb/utils/cache.py
+++ b/seismiqb/utils/cache.py
@@ -18,87 +18,138 @@ class _GlobalCacheClass:
     Note, the container saves only info for objects which use :class:`~.lru_cache` and :class:`~.cached_property`.
 
     TODO: proper removal instances control
+
+    cache_instances = {
+        'methods': {'name': cache_instance, ...},
+        'properties': {'name': {owner_instance, ...}, ...}
+    }
     """
     def __init__(self):
-        self.cached_objects = {'properties': set(), 'methods': set()} # Evaluated on the modules init
-        self.instances = set()
+        self.cache_instances = {'properties': {}, 'methods': {}} # Evaluated on the modules init
+
+    @staticmethod
+    def _get_method_length(cache_instance):
+        """ .. """
+        cache_length = 0
+        for cache in cache_instance.cache.values():
+            cache_length += len(cache)
+        return cache_length
+
+    @staticmethod
+    def _get_property_length(name, owners):
+        """ .. """
+        cache_length = 0
+        for owner in owners:
+            if name in owner.__dict__:
+                cache_length += 1
+        return cache_length
 
     @property
     def length(self):
         """ Total cache length. """
         cache_length = 0
-        for instance in self.instances:
-            cache_length += instance.cache_length
+        for cache_instance in self.cache_instances['methods'].values():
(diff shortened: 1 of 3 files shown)


seismiqb/utils/cache.py
```diff
@@ -185,65 +195,57 @@ class CacheMixin:
...
+        if objects == 'properties':
+            return self.cached_objects['properties'], set()
+        if objects == 'methods':
+            return set(), self.cached_objects['methods']
 
+        if not isinstance(objects, set):
+            objects = {objects}
+
+        properties = objects & self.cached_objects['properties']
+        methods = objects & self.cached_objects['methods']
```
SergeyTsimfer, line 236: Why do we need this functionality?

seismiqb/utils/cache.py
```diff
@@ -185,65 +195,57 @@ class CacheMixin:
     You can use this mixin for cache introspection and clearing cached data.
     """
     #pylint: disable=redefined-builtin
+    _global_cache_container = {'properties': set(), 'methods': set()}
```
SergeyTsimfer, line 199: I would argue for moving this container to `lru_cache` class itself

Also, I would add a method to get global cache size / cache repr: that would really help introspection

HollowPrincess: We use this container in `lru_cache`, `cached_property` and `CacheMixin`. I think the container must be separated from them all.

seismiqb/utils/cache.py
```diff
@@ -11,6 +12,166 @@
...
+
+        GlobalCache.cache_instances['properties'][self.name] = set()
+
+    def __set_name__(self, owner, name):
+        """ .. __set_name__ call for correct implementation"""
+        self.functools_func.__set_name__(owner, name)
+
+    def __get__(self, instance, owner=None):
+        """ .. save and get value"""
+        GlobalCache.cache_instances['properties'][self.name].add(instance)
```
SergeyTsimfer, line 172: this way, this reference to `instance` is kept in persistent `GlobalCache` object, therefore `instance` will never be dereferenced
Meaning that this is an endless memory leak

seismiqb/utils/cache.py
```diff
@@ -11,6 +12,166 @@
 import pandas as pd
 
 
+class _GlobalCacheClass:
```
SergeyTsimfer, line 15: Feels like this can be further simplified: current logic is complex and error-prone (and already leaks memory)

HollowPrincess:
Make cache in instances, try to make `GlobalCache` with weakref

seismiqb/utils/cache.py
```diff
@@ -65,24 +157,42 @@ def __init__(self, maxsize=128, attributes=None, apply_by_default=True, copy_on_
...
-            self.cache[instance_hash] = OrderedDict()
-            self.is_full[instance_hash] = False
             self.stats[instance_hash] = {'hit': 0, 'miss': 0}
 
-    def make_key(self, instance, args, kwargs):
+    def make_key(self, instance, func, args, kwargs):
         """ Create a key from a combination of method args and instance attributes. """
-        key = list(args[1:] if args[0] is instance else args)
+        # We get names and values for args and defaults from the func signature
+        func_signature = signature(func).parameters
```
SergeyTsimfer, line 171: Calling `signature` of a function at each function call is very slow

seismiqb/utils/cache.py
```diff
@@ -117,15 +227,24 @@ def wrapper(*args, **kwargs):
...
+            GlobalCache.instances_with_cache.add(instance)
+
+            key = self.make_key(instance, func, args, kwargs)
             instance_hash = self.compute_hash(instance)
 
             # If result is already in cache, just retrieve it and update its timings
             with self.lock:
-                result = self.cache[instance_hash].get(key, self.default)
+                cache = instance.cache[self.attrname]
+                result = cache.get(key, self.default)
```
SergeyTsimfer, line 243: it seems like `self.lock` should be used only for key replacement

seismiqb/utils/cache.py
```diff
@@ -135,28 +254,33 @@ def wrapper(*args, **kwargs):
...
+                    cache.popitem(last=False)
+                    cache[key] = result
                 else:
-                    self.cache[instance_hash][key] = result
-                    self.is_full[instance_hash] = (len(self.cache[instance_hash]) >= self.maxsize)
+                    cache[key] = result
+
             return copy(result) if copy_on_return else result
 
+        self.attrname = func.__name__ # used as a cache key in instances
```
SergeyTsimfer, line 268: `attrname` is not very descriptive

SergeyTsimfer: Also, why here you use `__name__`, but `__qualname__` as a key for `GlobalCache`?

seismiqb/utils/cache.py
```diff
@@ -182,137 +306,48 @@ def flatten_nested(iterable):
...
+            Attribute name. If None, then get total cache length.
         """
-        cached_properties, cached_methods = self.get_cached_objects(objects)
-
         cache_length_accumulator = 0
 
-        for property_name in cached_properties:
-            if property_name in self.__dict__:
-                cache_length_accumulator += 1
+        if hasattr(self, 'cache'):
```
SergeyTsimfer, line 321: it seems like you can make one method `get_cache_values` and re-use it in `get_cache_length` and `get_cache_size`.

Also, maybe we should use the word `size` as the amount of cached items and `nbytes` as the amount of memory taken?

seismiqb/utils/cache.py
```diff
@@ -1,21 +1,113 @@
...
+    """ Methods for global cache management.
+
+    Note, this class controls only objects which use :class:`~.lru_cache`.
+    So, for properties you need to use both `property` and `lru_cache` decorators for proper cache introspection.
+    """
+    def __init__(self):
+        """ Initialize containers with cache references and instances with cached objects.
+
+        Note, the `cache_references` container is filled on the modules import stage."""
+        self.cache_references = {} # for tests and debugging, helps to get cache info such as maxsize or stats
```
SergeyTsimfer, line 25: This contains a reference to one of the attributes of `instance`, so the `instance` is never deleted

seismiqb/utils/cache.py
```diff
@@ -182,137 +306,48 @@ def flatten_nested(iterable):
 class CacheMixin:
```
SergeyTsimfer, line 306: We can add methods to get total size of cache for all objects of the current class. Something like:
```python
@classmethod
def get_class_cache_size(cls):
    cache_size = 0
    for instance in GlobalCache.instances_with_cache:
        if isinstance(instance, cls):
            cache_size += instance.cache_size
    return cache_size
```
should do the job

seismiqb/utils/cache.py
HollowPrincess: Not true, there are links on objects which are created on init stage (`lru_cache` instance for each method). No cached object instances are needed (such horizons or geometries).

If it confuses, I can delete this object. I use it only in tests to get max cache size for methods.

seismiqb/utils/cache.py
HollowPrincess: Because `__name__` is a valid key for `instance.cache` dict when `__qualname__ ` is not.
Additionally, `__qualname__ ` is the only existing name for entity, when `__name__` can be not unique.

seismiqb/utils/cache.py
HollowPrincess: Also, I think it is more close to the `GlobalCache` than to the instance cache.
Do we need a level argument (with options 'global'/'total', 'classes', 'instances') in the `GobalCache` methods instead of this idea?

seismiqb/utils/cache.py
```diff
@@ -65,24 +214,46 @@ def __init__(self, maxsize=128, attributes=None, apply_by_default=True, copy_on_
...
+        args_and_defaults = [name for name in self.func_signature.keys()
+                             if (name not in kwargs.keys()) and (name != 'self')]
+
+        # Process default values
+        for default_param in args_and_defaults[len(args):]:
+            default_value = self.func_signature.get(default_param).default
+            args.append(default_value)
+
+        # Create key from args and defaults
+        key = list(zip(args_and_defaults, args))
```
SergeyTsimfer, line 247: what is the time cost of those checks?

seismiqb/utils/cache.py
```diff
@@ -117,15 +288,23 @@ def wrapper(*args, **kwargs):
...
+            key = self.make_key(instance, func, args, kwargs)
             instance_hash = self.compute_hash(instance)
 
             # If result is already in cache, just retrieve it and update its timings
-            with self.lock:
-                result = self.cache[instance_hash].get(key, self.default)
-                if result is not self.default:
-                    del self.cache[instance_hash][key]
-                    self.cache[instance_hash][key] = result
+            cache = instance.cache[self.cached_attr]
```
SergeyTsimfer, line 302: `instance_cache`, maybe?

seismiqb/utils/cache.py
```diff
@@ -135,28 +314,33 @@ def wrapper(*args, **kwargs):
...
             with self.lock:
                 self.stats[instance_hash]['miss'] += 1
-                if key in self.cache[instance_hash]:
+
+                if key in cache:
                     pass
-                elif self.is_full[instance_hash]:
-                    self.cache[instance_hash].popitem(last=False)
-                    self.cache[instance_hash][key] = result
+                elif len(cache) == self.maxsize:
```
SergeyTsimfer, line 320: I am fine with removing `is_full` dict, but keep the check `>=`. Just in case

seismiqb/utils/cache.py
```diff
@@ -1,21 +1,169 @@
...
+            if level == 'total':
+                result = pd.DataFrame(result, index=['total'])
+            elif level == 'class':
+                result = pd.DataFrame(result).T
+            elif level == 'instance':
+                result = pd.concat({k: pd.DataFrame(v).T for k, v in result.items()}, axis=0)
+
+        return result
+
+    def get_cache_repr(self, format='dict'):
```
SergeyTsimfer, line 105: Why is that function not a simple `repr` for the data, returned by `get_stats`, with the same parameters?

HollowPrincess: Because it processes arguments and I want to have a separate method.

PR #225, opened 2022-10-17
Fixes and updates (`SyntheticField`, `SeismicCropBatch`, `GeometryHDF5`, `nn.utils`)

HollowPrincess:
- Fixes in `SyntheticField`
 - Add `resize` for `SeismicCropBatch`
 - Add dead traces computation in ` GeometryHDF5` defaults
 - Add `show_slide_inference` method for models
 - Add `np.uint8` conversion into `Accumulator3D`

### DIFF
--- a/seismiqb/geometry/converted.py
+++ b/seismiqb/geometry/converted.py
@@ -82,21 +82,14 @@ class GeometryHDF5(Geometry):
         for key, value in kwargs.items():
             setattr(self, key, value)
 
-    def compute_dead_traces(self, frequency=100):
+    def compute_dead_traces(self):
         """ Fallback for dead traces matrix computation, if no full stats are collected. """
-        crop = None
+        slide = None
         self.dead_traces_matrix = np.zeros(shape=self.spatial_shape, dtype=np.bool_)
 
-        for idx in range(0, self.depth, frequency):
-            if idx + frequency > self.depth:
-                crop = None # buffer shape can be invalid for the last data chunk
-
-            locations = (slice(0, self.lengths[0]), slice(0, self.lengths[1]), slice(idx, idx+frequency))
-            crop = self.load_crop_native(locations=locations, axis=0, buffer=crop)
-
-            std_matrix = np.std(crop, axis=-1)
-
-            self.dead_traces_matrix |= (std_matrix == 0).astype(np.bool_)
+        for idx in range(self.shape[0]):
+            slide = self.load_slide_native(index=idx, axis=0, buffer=slide)
+            self.dead_traces_matrix[idx, :] = (np.std(slide, axis=-1) == 0).astype(np.bool_)
 
         self.n_dead_traces = np.sum(self.dead_traces_matrix)
         self.n_alive_traces = np.prod(self.lengths) - self.n_dead_traces
--- a/seismiqb/geometry/converted.py
+++ b/seismiqb/geometry/converted.py
@@ -82,18 +82,20 @@ class GeometryHDF5(Geometry):
         for key, value in kwargs.items():
             setattr(self, key, value)
 
-    def compute_dead_traces(self):
+    def compute_dead_traces(self, frequency=100):
         """ Fallback for dead traces matrix computation, if no full stats are collected. """
-        slide = None
-        self.dead_traces_matrix = np.zeros(shape=self.spatial_shape, dtype=np.bool_)
+        n_slides = self.depth // frequency + 1
+        buffer = np.empty((n_slides, *self.spatial_shape), dtype=self.dtype)
 
-        for idx in range(self.shape[0]):
-            slide = self.load_slide_native(index=idx, axis=0, buffer=slide)
-            self.dead_traces_matrix[idx, :] = (np.std(slide, axis=-1) == 0).astype(np.bool_)
+        for idx in range(0, self.depth, frequency):
+            buffer[idx//frequency, :, :] = self.load_slide_native(index=idx, axis=2,
+                                                                  buffer=buffer[idx//frequency, :, :])
 
+        std_matrix = np.std(buffer, axis=0)
+
+        self.dead_traces_matrix = (std_matrix == 0).astype(np.bool_)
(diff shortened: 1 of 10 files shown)


seismiqb/geometry/converted.py
```diff
@@ -78,9 +78,29 @@ def set_default_index_attributes(self, **kwargs):
...
+        """ Fallback for dead traces matrix computation, if no full stats are collected. """
+        crop = None
+        self.dead_traces_matrix = np.zeros(shape=self.spatial_shape, dtype=np.bool_)
+
+        for idx in range(0, self.depth, frequency):
+            if idx + frequency > self.depth:
+                crop = None # buffer shape can be invalid for the last data chunk
+
+            locations = (slice(0, self.lengths[0]), slice(0, self.lengths[1]), slice(idx, idx+frequency))
+            crop = self.load_crop_native(locations=locations, axis=0, buffer=crop)
```
SergeyTsimfer, line 95: Why not just `load_slide` to get data?

seismiqb/nn/utils.py
```diff
@@ -0,0 +1,143 @@
...
+import numpy as np
+from scipy.ndimage import find_objects, gaussian_filter
+from scipy.ndimage.morphology import binary_dilation
+from scipy.signal import find_peaks
+from skimage.measure import label
+
+from batchflow import Pipeline, B
+from batchflow.models.torch import TorchModel
+
+from seismiqb import SeismicDataset, RegularGrid, Accumulator3D
```
SergeyTsimfer, line 13: `from ..dataset import SeismicDataset`, etc

seismiqb/nn/utils.py
```diff
@@ -0,0 +1,143 @@
...
+    accumulator = Accumulator3D.from_grid(grid=grid, aggregation='weighted', fill_value=0)
+
+    # Inference
+    inference_pipeline = (
+        Pipeline()
+        .make_locations(generator=grid, batch_size=batch_size)
+        .load_cubes(dst='images')
+        .normalize(src='images')
+
+        .load_model(name='model', model_class=TorchModel, mode='dynamic', path=model_path)
```
SergeyTsimfer, line 84: this re-loads model on every iteration
why not use `model_or_path` as parameter, which can accept either ready-to-use instance of `TorchModel` or path to it?

seismiqb/nn/utils.py
```diff
@@ -0,0 +1,143 @@
...
+        .predict_model('model', inputs=B('images'), outputs='sigmoid', save_to=B('predictions'))
+        .update_accumulator(src='predictions', accumulator=accumulator)
+    ) << SeismicDataset(field)
+
+    inference_pipeline.run(n_iters=grid.n_iters, notifier='t')
```
SergeyTsimfer, line 89: add `pbar=False` parameter

seismiqb/nn/utils.py
```diff
@@ -0,0 +1,143 @@
...
+from batchflow.models.torch import TorchModel
+
+from seismiqb import SeismicDataset, RegularGrid, Accumulator3D
+
+
+def show_slide_inference(model_path, field, loc, axis=0, zoom='auto',
+                         batch_size=64, crop_shape=(1, 256, 512),
+                         inference_3d=True, inference_width=10,
+                         minsize=5, dilation_iterations=1,
+                         savepath=None, **kwargs):
```
SergeyTsimfer, line 20: add `plot` parameter to disable plotting

return `prediction_slide` data

and lets keep signature of this function as close to `field.load_slide/field.show_slide`, as possible: 
```python
def predict_slide(field, index, axis=0, model_or_path, ...)
```

maybe even split into two functions -- one for making prediction, second to make prediction + showing it

seismiqb/geometry/converted.py
SergeyTsimfer: It would load the same amount of data

seismiqb/geometry/converted.py
```diff
@@ -78,9 +78,22 @@ def set_default_index_attributes(self, **kwargs):
...
 
+    def compute_dead_traces(self):
+        """ Fallback for dead traces matrix computation, if no full stats are collected. """
+        slide = None
+        self.dead_traces_matrix = np.zeros(shape=self.spatial_shape, dtype=np.bool_)
+
+        for idx in range(self.shape[0]):
+            slide = self.load_slide_native(index=idx, axis=0, buffer=slide)
+            self.dead_traces_matrix[idx, :] = (np.std(slide, axis=-1) == 0).astype(np.bool_)
+
```
SergeyTsimfer, line 93: that looks really strange

why don't you use `load_slide_native(axis=2)`?

seismiqb/geometry/converted.py
```diff
@@ -78,9 +78,24 @@ def set_default_index_attributes(self, **kwargs):
...
         self.n_traces = np.prod(self.shape[:2])
         self.delay, self.sample_rate = 0.0, 1.0
+        self.compute_dead_traces()
         for key, value in kwargs.items():
             setattr(self, key, value)
 
+    def compute_dead_traces(self, frequency=100):
+        """ Fallback for dead traces matrix computation, if no full stats are collected. """
+        n_slides = self.depth // frequency + 1
+        buffer = np.empty((n_slides, *self.spatial_shape), dtype=self.dtype)
```
SergeyTsimfer, line 88: do you really need it?
is it even used in case of public `h5py` calls?

seismiqb/geometry/converted.py
```diff
@@ -78,9 +78,24 @@ def set_default_index_attributes(self, **kwargs):
...
+
+        for idx in range(0, self.depth, frequency):
+            buffer[idx//frequency, :, :] = self.load_slide_native(index=idx, axis=2,
```
SergeyTsimfer, line 91: If you really wanted to make it efficient, you should have used `Accumulator2D` with `std` aggregation, same as for metrics computation

I don't think it is needed here, and you can just load all slides in a list and get `std` along the axis

SergeyTsimfer (approved):

PR #240, opened 2022-12-29
New extractor for Faults

HollowPrincess:


seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,578 @@
...
+# TODO: add info about connected prototypes (maybe into FaultPrototype)
+
+class FaultExtractor:
+    """ ..!!..
+
+    Main logic implementation. There are much to do and much to optimize in this code.
+
+    A few comments about the main idea:
+    We extract connected components on each slide.
+    We take unmerged component and find the closest component on next slide.
```
AlexeyKozhevin, line 24: How to choose that component?

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,578 @@
...
+    Merge them into a prototype (3d points body).
+    Repeat this for the last merged component while can.
```
AlexeyKozhevin, line 26: How to choose next components to start prototype collecting?

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,578 @@
...
+    - Prototype is a 3d points body of merged components.
+    """
+    def __init__(self, skeletonized_array, smoothed_array, orientation=0, component_len_threshold=0):
+        """ Init data container with components info for each slide.
+
+        ..!!..
+
+        Parameters
+        ----------
+        orientation : {0, 1}
```
AlexeyKozhevin, line 41: Let's use `direction` in that sense

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,578 @@
...
+            for idx, object_bbox in enumerate(objects):
+                # Refined coords: we refine skeletonize effects by applying it on limited area
+                dilation_axis = self.orthogonal_orientation
+                dilation_ranges = (np.clip(object_bbox[0].start - self.dilation // 2, 0, None),
+                                   np.clip(object_bbox[0].stop + self.dilation // 2, 0, self.shape[dilation_axis]))
+
+                object_mask = labeled[dilation_ranges[0]:dilation_ranges[1], object_bbox[-1]] == idx + 1
+                object_mask = binary_dilation(object_mask, structure=np.ones((1, self.dilation), bool))
+
+                dilated_coords_2d = np.nonzero(object_mask)
```
AlexeyKozhevin, line 83: `dilated_coords_2d` can be created directly from `np.nonzero(labeled[object_bbox])`. Just add +1/-1 to the column

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,578 @@
...
+
+            self.container[start_slide_idx]['mergeable'][idx] = False
+
+            component = self.container[start_slide_idx]['objects_coords'][idx]
+            prototype = component
+        else:
+            prototype = self.prototypes_queue.popleft()
+
+            start_slide_idx = np.max(prototype[:, self.orientation])
+            component = prototype[prototype[:, self.orientation] == start_slide_idx]
```
AlexeyKozhevin, line 195: May be you should store coordinates slide-by-slide to avoid such slicing?

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,578 @@
...
+                    closest_component = current_component
+                    closest_component_bbox = current_bbox
+
+                    if min_distance == 0:
+                        break
+
+        if closest_component is not None:
+            intersection_height = intersection_borders[1] - intersection_borders[0]
+
+            if (intersection_height < 0.2*len(component)/self.dilation) or \
```
AlexeyKozhevin: Hardcode?

HollowPrincess: Deleted in the revious commit

seismiqb/labels/fault/extractor.py
HollowPrincess: We need **unique** coordinates and current implemetation is faster than using `dilate_coords` (which makes +1/-1 to the column and get only unique coordinates). You can see the implementation in the PR file `utils.py`.

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,578 @@
...
+    def _find_not_merged_component(self):
+        """ Find the longest not merged item on the minimal slide. """
+        # TODO: add info about slides with mergeable components and reduce cycle iterations amount
+        idx = None
+
+        for slide_idx in range(self.shape[self.orientation]):
+            slide_info = self.container[slide_idx]
+
+            if slide_info['mergeable'].any():
+                max_len = np.max(slide_info['lengths'][slide_info['mergeable']])
```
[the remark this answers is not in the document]
HollowPrincess: Thanks, that looks much better

seismiqb/labels/fault/utils.py
```diff
@@ -0,0 +1,164 @@
...
+
+@njit
+def restore_coords_from_projection(coords, buffer, axis):
+    """ Find `axis` coordinates from coordinates and their projection.
+
+    ..!!..
+    buffer : np.ndarray
+        Buffer with projection coordinates.
+    """
+    known_axes = [i for i in range(3) if i != axis]
```
SergeyTsimfer, line 158: numba is not very happy with that list

SergeyTsimfer: nut sure if the numba is even speeding up anything in this function

seismiqb/labels/fault/utils.py
```diff
@@ -0,0 +1,164 @@
...
+    contour_coords[:, 2] = coords_2d[1] + origin[1]
+    return contour_coords
+
+@njit
+def restore_coords_from_projection(coords, buffer, axis):
+    """ Find `axis` coordinates from coordinates and their projection.
+
+    ..!!..
+    buffer : np.ndarray
+        Buffer with projection coordinates.
```
SergeyTsimfer, line 156: specifying shapes in docs would help a lot

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,610 @@
...
+        ----------
+        direction : {0, 1}
+            Prediction direction corresponding ilines (0) or xlines (1).
+        component_len_threshold : int
+            Threshold to filter out too small connected components on data slides.
+        """
+        self.shape = skeletonized_array.shape
+
+        self.direction = direction
+        self.orthogonal_direction = int(not self.direction)
```
SergeyTsimfer, line 50: Negating integers is `not` a great idea. Why not `1 - self.direction`, as everywhere else?

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,610 @@
...
+            labeled = connected_components(mask > 0)
+            objects = find_objects(labeled)
+
+            # Get components info
+            coords, bboxes, lengths = [], [], []
+
+            for idx, object_bbox in enumerate(objects):
+                # Refined coords: we refine skeletonize effects by applying it on limited area
+                dilation_axis = self.orthogonal_direction
+                dilation_ranges = (np.clip(object_bbox[0].start - self.dilation // 2, 0, None),
```
SergeyTsimfer, line 76: i would avoid `np.clip` here: 
```python
max(object_bbox[0].start - self.dilation // 2, 0),
min(object_bbox[0].stop  + self.dilation // 2, self.shape[dilation_axis])
```

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,610 @@
...
+            smoothed = np.take(smoothed_array, slide_idx, axis=self.direction)
+
+            # Extract connected components from the slide
+            labeled = connected_components(mask > 0)
+            objects = find_objects(labeled)
+
+            # Get components info
+            coords, bboxes, lengths = [], [], []
+
+            for idx, object_bbox in enumerate(objects):
```
SergeyTsimfer, line 73: use `enumerate(..., start=1)` -- this way you would not need to use `i+1` everywhere and can use just `i`

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,610 @@
...
+                # Refined coords: we refine skeletonize effects by applying it on limited area
+                dilation_axis = self.orthogonal_direction
+                dilation_ranges = (np.clip(object_bbox[0].start - self.dilation // 2, 0, None),
+                                   np.clip(object_bbox[0].stop + self.dilation // 2, 0, self.shape[dilation_axis]))
+
+                object_mask = labeled[dilation_ranges[0]:dilation_ranges[1], object_bbox[-1]] == idx + 1
+                object_mask = binary_dilation(object_mask, structure=np.ones((1, self.dilation), bool))
```
SergeyTsimfer, line 80: create dilation kernel once
also, `cv.dilate` may be faster

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,610 @@
...
+
+                dilated_coords[:, self.direction] = slide_idx
+                dilated_coords[:, dilation_axis] = dilated_coords_2d[0] + dilation_ranges[0]
+                dilated_coords[:, 2] = dilated_coords_2d[1] + object_bbox[1].start
+
+                smoothed_values = smoothed[dilated_coords[:, dilation_axis], dilated_coords[:, -1]]
+
+                refined_coords = thin_coords(coords=dilated_coords, values=smoothed_values)
+
+                coords.append(refined_coords)
```
SergeyTsimfer, line 94: is there actually a difference between using original coords and those refined ones?

SergeyTsimfer: Also, that is almost the same, as dilating the entire slide (`mask`) and extract components from it

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,610 @@
...
+                dilation_axis = self.orthogonal_direction
+                dilation_ranges = (np.clip(object_bbox[0].start - self.dilation // 2, 0, None),
+                                   np.clip(object_bbox[0].stop + self.dilation // 2, 0, self.shape[dilation_axis]))
+
+                object_mask = labeled[dilation_ranges[0]:dilation_ranges[1], object_bbox[-1]] == idx + 1
+                object_mask = binary_dilation(object_mask, structure=np.ones((1, self.dilation), bool))
+
+                dilated_coords_2d = np.nonzero(object_mask)
+
+                dilated_coords = np.zeros((len(dilated_coords_2d[0]), 3), dtype=int)
```
SergeyTsimfer, line 84: do you actually need `np.int64` dtype here?

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,610 @@
...
+                refined_coords = thin_coords(coords=dilated_coords, values=smoothed_values)
+
+                coords.append(refined_coords)
+
+                # Bbox
+                bbox = np.empty((3, 2), int)
+
+                bbox[self.direction, :] = slide_idx
+                bbox[self.orthogonal_direction, :] = (np.min(refined_coords[:, self.orthogonal_direction]),
+                                                      np.max(refined_coords[:, self.orthogonal_direction]))
```
SergeyTsimfer, line 101: you dont need to take `min` of an array: you already know the bbox (`object_bbox` variable` and can subtract/add dilation on either side

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,610 @@
...
+
+                bboxes.append(bbox)
+
+                # Length
+                lengths.append(len(refined_coords))
+
+            # Filter components by length
+            lengths = np.array(lengths)
+            is_too_small = lengths <= component_len_threshold
+            lengths[is_too_small] = -1 # -1 is a flag for unmergeable components
```
SergeyTsimfer, line 112: if those components are never used, you can do that filtration before expensive operations with coords

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,610 @@
...
+                'bboxes': bboxes,
+                'lengths': lengths
+            }
+
+        self._first_slide_with_mergeable = 0
+
+    # Prototypes extraction
+    def extract_prototypes(self):
+        """ Extract all fault prototypes from the point cloud. """
+        prototype = self.extract_prototype()
```
SergeyTsimfer, line 125: `prototype = None`

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,610 @@
...
+                    new_prototype = prototype.split(split_depth=prototype_split_indices[1], cut_upper_part=False)
+                    self.prototypes_queue.append(new_prototype)
+
+                prototype.append(component, slide_idx=slide_idx_)
+            else:
+                break
+
+        return prototype
+
+    def _find_not_merged_component(self):
```
SergeyTsimfer, line 174: maybe, add a length threshold to use only big components at first?

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,610 @@
...
+
+    def _find_closest_component(self, component, slide_idx, distances_threshold=5,
+                                depth_iteration_step=10, depths_threshold=5):
+        """ Find the closest component to component on the slide, get splitting depths for them if needed.
+
+        ..!!..
+        """
+        # Process inputs
+        # Dilate component bbox for detecting close components: component on next slide can be shifted
+        component_bbox = np.column_stack([np.min(component, axis=0), np.max(component, axis=0)])
```
SergeyTsimfer, line 196: dont you know that bbox from `self.container`?

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,610 @@
...
+                current_component = self.container[slide_idx]['coords'][idx]
+
+                intersection_depths = (max(component_bbox[-1, 0], current_bbox[-1, 0]),
+                                       min(component_bbox[-1, 1], current_bbox[-1, 1]))
+
+                step = np.clip((intersection_depths[1]-intersection_depths[0])//3, 1, depth_iteration_step)
+
+                distance = max_depthwise_distance(component, current_component,
+                                                  depths_ranges=intersection_depths, step=step,
+                                                  axis=self.orthogonal_direction, max_threshold=min_distance)
```
SergeyTsimfer, line 225: random thought: transposing slides to `(depth, x)` orientation prior to labeling would make `depth` the first axis in `coords`, which would make it sorted. This can save you `log(N)` time, because having sorted array in most of your `utils.py` functions would speed up them a lot

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,610 @@
...
+        component_bbox[self.orthogonal_direction, 1] += self.dilation // 2
+
+        min_distance = distances_threshold
+
+        # Init returned values
+        closest_component = None
+        prototype_split_indices = [None, None]
+
+        # Iter over components and find the closest one
+        for idx, current_bbox in enumerate(self.container[slide_idx]['bboxes']):
```
SergeyTsimfer, line 207: i would rather use `other_bbox` naming to name the component candidate
`current` seems to be the one, that you are trying to merge with

seismiqb/labels/fault/extractor.py
SergeyTsimfer: random idea №2: you could save `centroid` for each of the component, and compute distance to it. Should be faster and even more accurate

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,610 @@
...
+
+                if adjoining_borders is None:
+                    continue
+
+                # Check that bboxes overlap is enough
+                intersection_threshold = min(prototype_1.bbox[overlapping_axis, 1]-prototype_1.bbox[overlapping_axis, 0],
+                                             prototype_2.bbox[overlapping_axis, 1]-prototype_2.bbox[overlapping_axis, 0])
+                intersection_threshold *= intersection_ratio_threshold
+
+                intersection_length = adjoining_borders[overlapping_axis][0] - adjoining_borders[overlapping_axis][1]
```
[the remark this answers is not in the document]
SergeyTsimfer: `adjoint_check`, `intersection_check`, `contour_check`

seismiqb/labels/fault/utils.py
```diff
@@ -0,0 +1,164 @@
...
+    """ Check bboxes intersection on axes. """
+    for axis in axes:
+        borders_delta = min(bbox_1[axis, 1], bbox_2[axis, 1]) - max(bbox_1[axis, 0], bbox_2[axis, 0])
+
+        if borders_delta < 0:
+            return False
+    return True
+
+@njit
+def bboxes_adjoining(bbox_1, bbox_2):
```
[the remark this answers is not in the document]
HollowPrincess: Concat prototypes timings:

Version with list: 1min 14s
Version with np.array: 1min 16s

2s can be a calculation error, added version with numpy array, because of suitable memory preallocation.

seismiqb/labels/fault/utils.py
```diff
@@ -0,0 +1,164 @@
...
+            previous_value = current_value
+
+    # last depth update
+    output[position, :] = point_to_save
+    position += 1
+
+    return output[:position, :]
+
+# Distance evaluation
+def bboxes_intersected(bbox_1, bbox_2, axes=(0, 1, 2)):
```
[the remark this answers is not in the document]
HollowPrincess: Timings for prototypes extraction:

current: 42.1 s
njit: 30.9 s

Will apply njit. Earlier wasn't faster, strange... Thanks for the comment.

seismiqb/labels/fault/utils.py
```diff
@@ -0,0 +1,164 @@
...
+    # Make 2d projection on projection_axis
+    bbox = np.column_stack([np.min(coords, axis=0), np.max(coords, axis=0)])
+    bbox = np.delete(bbox, projection_axis, 0)
+
+    # Create object mask
+    origin = bbox[:, 0]
+    image_shape = bbox[:, 1] - bbox[:, 0] + 1
+
+    mask = np.zeros(image_shape, bool)
+    mask[coords[:, anchor_axis] - origin[0], coords[:, 2] - origin[1]] = 1
```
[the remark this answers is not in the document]
HollowPrincess: Prototypes concatenation:

Base timings: 2.06 s
Timings for your idea: 2.21 s

seismiqb/labels/fault/utils.py
HollowPrincess: Timings for prototypes concatenation:

Base: 1min 14s

Without njit: 1min 20s
With tuple instead of list (without njit): 1min 18s

With np.array instead of list: 1min 12s
With np.array instead of list (without njit):  1min 20s

seismiqb/labels/fault/extractor.py
HollowPrincess: Extractor init timings (for two different data volumes):

Current: 10.4 s and 1min 1s
cv.dilate for mask (results are different because dilated components are overlapped): 9.87 s and 1min 2s
cv.dilate for each component: 9.59 s and 59 s

seismiqb/labels/fault/extractor.py
HollowPrincess: 1) Yes, there is a difference: peaking lose some points in difficult areas and peaking in limited areas creates more accurate components. I ran code with and without refined coords and got different results.

2) No, not the same: dilated areas are intersected and mask dilation can make wrong dilated areas for some components. I checked, results are different. Time changes are not very different (look at the previous comment) and this update isn't worth the accuracy.

seismiqb/labels/fault/extractor.py
HollowPrincess: 👍

seismiqb/labels/fault/utils.py
```diff
@@ -0,0 +1,164 @@
...
+    # Get object contour
+    contour = mask ^ binary_erosion(mask)
+
+    # Extract coordinates from contour mask
+    coords_2d = np.nonzero(contour)
+
+    contour_coords = np.zeros((len(coords_2d[0]), 3), dtype=int)
+
+    contour_coords[:, anchor_axis] = coords_2d[0] + origin[0]
+    contour_coords[:, 2] = coords_2d[1] + origin[1]
```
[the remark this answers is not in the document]
HollowPrincess: Timings for prototypes concat:

Base :  2.08 s
Find contours: 1.51 s

*Timings can be differ from other because other logic has a few fixes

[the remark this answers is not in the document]
HollowPrincess: Timings for prototypes concat:

Base: 1min 10s
findContours: 1min 12s

seismiqb/labels/fault/extractor.py
HollowPrincess: Why?

seismiqb/labels/fault/extractor.py
HollowPrincess: I want to find components from left to right.
The biggest component can be from the middle of the surface and this not seems good.
So, I find the left slide with unmerged components and find the biggest component on it.

HollowPrincess:
TODO: check contours after prototypes concat

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,610 @@
...
+        overlapping_axis = self.direction if axis in (-1, 2) else 2
+
+        if axis in (-1, 2): # not necessary object borders for contour finding
+            removed_borders = ('up', 'down')
+        else:
+            removed_borders = ('left', 'right')
+
+        for j, prototype_1 in enumerate(self.prototypes):
+            for k, prototype_2 in enumerate(self.prototypes[j+1:]):
+                adjoining_borders = bboxes_adjoining(prototype_1.bbox, prototype_2.bbox)
```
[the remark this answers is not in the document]
HollowPrincess: Base timing: 2.72 s
With vectorization:  9.17 s

HollowPrincess:
TODO: presort bboxes and early stop for prototypes concat

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,681 @@
...
+    """ Add item_i and item_j to dependencies graph of elements to concat.
+
+    `to_concat` is the dict in the format {'owner_idx': [items_indices]} and contains
+    which components to concat into owner-component.
+    `concated_with` is the dict in the format {'item_idx': owner_idx} and contains
+    to which component (owner) merge item.
+
+    ..!!..
+    """
+    if item_i not in concated_with.keys():
```
SergeyTsimfer, line 645: `.keys()` is not needed

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,681 @@
...
+        to_concat = self.find_connected_prototypes(axis=axis)
+        remove_elements = []
+
+        for where_to_concat_idx, what_to_concat_indices in to_concat.items():
+            main_prototype = self.prototypes[where_to_concat_idx]
+
+            for idx in what_to_concat_indices:
+                main_prototype.concat(self.prototypes[idx])
+
+            remove_elements.extend(what_to_concat_indices)
```
SergeyTsimfer, line 338: I feel like you can massively speed up this thing by merging all prototypes `([where_to_concat_idx] + what_to_concat_indices)` at once: that would require ~3 lines of code changed in `Prototype.concat` to allow for more than 2 prototypes at once

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,681 @@
...
+                # Bbox
+                bbox = np.empty((3, 2), int)
+
+                bbox[self.direction, :] = slide_idx
+                bbox[self.orthogonal_direction, :] = dilation_ranges
+                bbox[-1, :] = object_bbox[-1].start, object_bbox[-1].stop - 1
+
+                bboxes.append(bbox)
+
+            # Filter components by length
```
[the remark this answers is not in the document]
HollowPrincess: It works, but now I apply no filtering, because I want to see the complete picture of what is happening.

seismiqb/labels/fault/extractor.py
HollowPrincess: Not faster:
Base timing: 16.9 s
For concat all at once: 17.2 s

HollowPrincess: Also, earlier I proposed the idea to concat prototypes at the moment and break the internal cycle. It is slower: 20.8 s

seismiqb/labels/fault/extractor.py
HollowPrincess: Centroid doesn'twork because component on next slide can be much longer and its' centroid is much lower.
Example:
Component 1 is green, component 2 is red.
Red circles - approximate centroids (prefic - compoent idx, suffx - slide idx).
Image one is slide number x, image 2 is slide x+1.
The green component on slide x+1 is the closest component to green component on slide x. he same for the red component.
As you can see, distance between C1_x and C2_x+1 is smaller than for C1_x and C1_x+1 which is not good for our case.
Снимок экрана от 2023-01-30 15-40-56

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,966 @@
...
+        self.container = {}
+
+        self.prototypes_queue = deque() # prototypes for extension
+        self.prototypes = [] # extracted prototypes
+
+        # Process data slides: extract connected components and their info
+        for slide_idx in Notifier('t')(range(self.shape[self.direction])):
+            # Get smoothed data slide
+            if is_smoothed_is_np:
+                smoothed = np.take(smoothed_data, slide_idx, axis=self.direction)
```
SergeyTsimfer, line 74: There is a method `take` of a `np.ndarray`s
We can add alias `take=load_slide` to Geometry so that look the same

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,966 @@
...
+                    continue
+
+                # Check closeness of some points (iter over intersection depths with some step)
+                # Faster then component intersection, but not so accurate
+                other_component = self.container[slide_idx]['coords'][idx]
+
+                intersection_depths = (max(component_bbox[-1, 0], other_bbox[-1, 0]),
+                                       min(component_bbox[-1, 1], other_bbox[-1, 1]))
+
+                step = np.clip((intersection_depths[1]-intersection_depths[0])//3, 1, depth_iteration_step)
```
SergeyTsimfer, line 269: i would rather use min/max functions here

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,966 @@
...
+                    min_distance = components_distances[1]
+                    intersection_borders = intersection_depths
+                    closest_component = other_component
+                    closest_component_bbox = other_bbox
+
+                    if min_distance == 0:
+                        # The closest component is founded
+                        break
+
+        if closest_component is not None:
```
SergeyTsimfer, line 292: Maybe, should be moved to a separate function? Otherwise, this method both finds and merges the closest component

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,966 @@
...
+        Parameters
+        ----------
+        smoothed_data : np.ndarray or :class:`~.Geometry` instance
+            A 3d array with smoothed predictions with shape corresponds to the field shape.
+        direction : {0, 1}
+            Prediction direction, can be 0 (ilines) or 1 (xlines).
+        skeletonized_data : np.ndarray or :class:`~.Geometry` instance, optional
+            A 3d array with skeletonized predictions with shape corresponds to the field shape.
+            If None is provided, then it will be evaluated.
+        component_len_threshold : int
```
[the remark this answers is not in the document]
HollowPrincess: No.
It is a threshold for length of the component.

Your naming can mean component size, prototype size.

I prefer the `length` word, because it is amount of component points (`len(coords)`).

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,966 @@
...
+    Repeat this for the last merged component while can.
+    In this way we extract all prototypes.
+    After that we find connected prototypes and concat them.
+
+    Main naming rules:
+    - Component is a 2d connected component on some slide.
+    - Prototype is a 3d points body of merged components.
+    """
+    def __init__(self, smoothed_data, direction, skeletonized_data=None, component_len_threshold=3):
+        """ Init data container with components info for each slide.
```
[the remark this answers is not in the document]
HollowPrincess: Ok, but I will write the whole class documentation later.

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,966 @@
...
+
+        Parameters
+        ----------
+        smoothed_data : np.ndarray or :class:`~.Geometry` instance
```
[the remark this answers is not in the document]
HollowPrincess: Sounds reasonable

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,966 @@
...
+            A 3d array with smoothed predictions with shape corresponds to the field shape.
+        direction : {0, 1}
+            Prediction direction, can be 0 (ilines) or 1 (xlines).
+        skeletonized_data : np.ndarray or :class:`~.Geometry` instance, optional
```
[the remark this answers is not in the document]
HollowPrincess: Only for speed up tests

seismiqb/labels/fault/extractor.py
HollowPrincess: It doesn't merge a component. It finds the closest component and split it if needed. Merging is `prototype.append(component, bbox=component_bbox, slide_idx=slide_idx_)` in the `.extract_prototype`

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,981 @@
...
+        self.shape = data.shape
+
+        self.direction = direction
+        self.orthogonal_direction = 1 - self.direction
+
+        self.dilation = 3 # constant for internal operations
+        self.component_len_threshold = component_len_threshold
+
+        self.container = self._init_container(data=data)
+        self._first_slide_with_mergeable = 0 # variable for internal operations speed up
```
SergeyTsimfer, line 57: `component_len_threshold` / `_first_slide_with_mergeable ` naming could be improved
`size_threshold` / `_unprocessed_slide_idx`?

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,981 @@
...
+                length = np.count_nonzero(object_mask)
+
+                if length <= self.component_len_threshold:
+                    continue
+
+                # Extract refined component coordinates
+                # skeletonize on the full slide is not so accurate than for limited area
+                object_mask = dilate(object_mask.astype(np.uint8), dilation_structure)
+
+                dilated_coords_2d = np.nonzero(object_mask)
```
SergeyTsimfer, line 101: what is the dtype of this object?

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,981 @@
...
+
+    # Prototypes extraction
+    def extract_prototypes(self):
+        """ Extract all fault prototypes from the point cloud. """
+        prototype = self.extract_prototype()
+
+        while prototype is not None:
+            self.prototypes.append(prototype)
+            prototype = self.extract_prototype()
+
```
SergeyTsimfer, line 150: I would rather prefer:
```python
while True:
    prototype = self.extract_prototype()
    if prototype is None:
        break
    self.prototypes.append(prototype)
```
but you do you

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,981 @@
...
+    def extract_prototype(self):
+        """ Extract one fault prototype from the point cloud. """
+        if len(self.prototypes_queue) == 0:
+            start_slide_idx, idx = self._find_not_merged_component()
+
+            if idx is None: # No components to concat
+                return None
+
+            component = self.container[start_slide_idx]['coords'][idx]
```
SergeyTsimfer, line 161: i would rather name it `component_coords`

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,981 @@
...
+        while prototype is not None:
+            self.prototypes.append(prototype)
+            prototype = self.extract_prototype()
+
+        return self.prototypes
+
+    def extract_prototype(self):
+        """ Extract one fault prototype from the point cloud. """
+        if len(self.prototypes_queue) == 0:
+            start_slide_idx, idx = self._find_not_merged_component()
```
SergeyTsimfer, line 156: `_find_next_component`?

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,981 @@
...
+            Depth-length threshold to decide to split closest component or prototype.
+            If one component is longer than another more than on depths_threshold,
+            then we need to split the longest one into parts:
+             - one part corresponds to the closest component;
+             - another correspond to the different component, which is not allowed for merge to the current prototype.
+        """
+        # Dilate component bbox for detecting close components: component on next slide can be shifted
+        dilated_bbox = component_bbox[self.orthogonal_direction, :] + (-self.dilation // 2, self.dilation // 2)
+        component_bbox[self.orthogonal_direction, 0] = max(0, dilated_bbox[0])
+        component_bbox[self.orthogonal_direction, 1] = min(dilated_bbox[1], self.shape[self.orthogonal_direction])
```
SergeyTsimfer, line 235:
```python
component_bbox[self.orthogonal_direction, 0] = max(component_bbox[self.orthogonal_direction, 0] - self.dilation // 2, 
                                                   0)
component_bbox[self.orthogonal_direction, 1] = min(component_bbox[self.orthogonal_direction, 1] + self.dilation // 2,
                                                   self.shape[self.orthogonal_direction])
```
it is not the same as was here before, is it?

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,981 @@
...
+            if self.container[slide_idx]['lengths'][idx] != -1:
+                # Check bboxes intersection
+                if not bboxes_intersected(component_bbox, other_bbox, axes=(self.orthogonal_direction, -1)):
+                    continue
+
+                # Check closeness of some points (iter over intersection depths with some step)
+                # Faster then component intersection, but not so accurate
+                other_component = self.container[slide_idx]['coords'][idx]
+
+                intersection_depths = (max(component_bbox[-1, 0], other_bbox[-1, 0]),
```
SergeyTsimfer, line 257: I would rather use word `overlap`, here and in other places, as it is already used in our library for this exact meaning

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,981 @@
...
+                components_distances = min_max_depthwise_distances(component, other_component,
+                                                                   depths_ranges=intersection_depths, step=step,
+                                                                   axis=self.orthogonal_direction,
+                                                                   max_threshold=min_distance)
+
+                if (components_distances[0] is None) or (components_distances[0] > 1):
+                    # Components are not close
+                    continue
+
+                if components_distances[1] < min_distance:
```
SergeyTsimfer, line 272:
```python
if components_distances[1] >= min_distance:
    continue
```
maybe?

SergeyTsimfer (commented):
the part that I've looked at is much cleaner and overall neat: pretty much all of the comments are naming/docs related
👍 

look at the rest of the code later

seismiqb/labels/fault/extractor.py
HollowPrincess: I agree with the second renaming, but not with the first.
`size_threshold` doesn't provide the information about size of what: component or prototype.

seismiqb/labels/fault/extractor.py
HollowPrincess: I don't like cycles which starts with `while True` =(

seismiqb/labels/fault/extractor.py
HollowPrincess: uint8

seismiqb/labels/fault/extractor.py
HollowPrincess: I changed code, this can not be applied

seismiqb/labels/fault/coords_utils.py
```diff
@@ -0,0 +1,271 @@
...
+    for depth in range(depths_ranges[0], depths_ranges[1]+1, step):
+        # Find distance for the current depth
+        coords_1_depth_slice = coords_1[coords_1[:, -1] == depth, axis]
+        coords_2_depth_slice = coords_2[coords_2[:, -1] == depth, axis]
+
+        distance = np.abs(coords_1_depth_slice[len(coords_1_depth_slice)//2] - \
+                          coords_2_depth_slice[len(coords_2_depth_slice)//2])
+
+        # Result processing
+        if (max_threshold is not None) and (distance >= max_threshold):
```
SergeyTsimfer, line 199: also, by setting `max_threshold` as required arg or with some big default value you can avoid the first condition, which is a free speed-up

seismiqb/labels/fault/coords_utils.py
```diff
@@ -0,0 +1,271 @@
...
+    ----------
+    coords_1, coords_2 : np.ndarrays of (N, 3) shape
+        Coords for which find axis-wise distances.
+    depths_ranges : sequence of two int values
+        Depth ranges along which to find distances.
+        Recommended to provide depth intersection ranges for coords.
+    step : int
+        Depth step to find distances.
+        For speed up we can find distances not for all depths.
+        For results accuracy, provide step = 1.
```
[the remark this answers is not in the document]
SergeyTsimfer: If you go this route, you can also check the `shapely` performance on the same task: it has both `distance` and `hausdorff_distance`

seismiqb/labels/fault/coords_utils.py
```diff
@@ -0,0 +1,271 @@
...
+    ----------
+    coords : np.ndarray of (N, 3) shape
+        3D object coordinates.
+    projection_axis : {0, 1}
+        Axis for making 2d projection for which we find contour.
+        Note, this function doesn't work for axis = 2.
+    """
+    # Make 2d projection
+    bbox = np.column_stack([np.min(coords, axis=0), np.max(coords, axis=0)])
+    bbox = np.delete(bbox, projection_axis, 0)
```
SergeyTsimfer, line 225: Why not use indexing instead?

seismiqb/labels/fault/coords_utils.py
```diff
@@ -0,0 +1,271 @@
...
+    mask = np.zeros(image_shape, np.uint8)
+    mask[coords[:, 1 - projection_axis] - origin[0], coords[:, 2] - origin[1]] = 1
+
+    # Get only the main object contour: object can contain holes with their own contours
+    contours, _ = cv.findContours(mask, cv.RETR_TREE, cv.CHAIN_APPROX_NONE)
+
+    # Extract unique and sorted coords
+    contour = contours[0].reshape(len(contours[0]), 2) # Can be non-unique
+
+    contour_coords = np.zeros((len(contour), 3), np.int32)
```
SergeyTsimfer, line 240: looks like `empty` will be faster here (but you would need to manually put zeros at projection axis`

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,1056 @@
...
+
+from .base import Fault
+from .postprocessing import skeletonize
+from .coords_utils import (bboxes_adjacent, bboxes_embedded, bboxes_intersected, dilate_coords, find_contour,
+                           thin_coords, min_max_depthwise_distances, restore_coords_from_projection)
+from ...utils import groupby_min, groupby_max
+
+
+class FaultExtractor:
+    """ Extract fault surfaces from array with fault labels (smoothed probabilities).
```
SergeyTsimfer, line 19: Extract fault surfaces from an array with predicted fault probability

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,1056 @@
...
+
+    We assume that connected 2d components owned by one prototype are looks the same.
```
SergeyTsimfer, line 21: What did you mean by that?

seismiqb/labels/fault/extractor.py
SergeyTsimfer: Removing the note about conventions used in docs (`component is 2d connected on a slide, prototype is ...`) makes things harder to read

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,1056 @@
...
+    data : np.ndarray or :class:`~.Geometry` instance
+        A 3d array with smoothed predictions with shape corresponds to the field shape.
+    direction : {0, 1}
+        Prediction direction, can be 0 (ilines) or 1 (xlines).
+    component_len_threshold : int
+        Threshold to filter out too small connected components on data slides.
+        If 0, then no filter applied (recommended for higher accuracy).
+        If more than 0, then extraction will be faster.
+    """
+    def __init__(self, data, direction, component_len_threshold=0):
```
SergeyTsimfer, line 84: should not there be `origin` as well? Otherwise, you always have to inference the whole field

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,1056 @@
...
+                # Previous length and len(refined_coords) are different
+                # because original coords have more than one coord per depth when refined are not
+                length = len(refined_coords)
+
+                if length <= self.component_len_threshold:
+                    continue
+
+                lengths.append(length)
+
+                probas = np.round(probas * 255).astype(refined_coords.dtype)
```
SergeyTsimfer, line 160: Why the multiplication?

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,1056 @@
...
+
+                dilated_coords = np.zeros((len(dilated_coords_2d[0]), 3), dtype=np.int32)
+
+                dilated_coords[:, self.direction] = slide_idx
+                dilated_coords[:, dilation_axis] = dilated_coords_2d[0] + dilation_ranges[0]
+                dilated_coords[:, 2] = dilated_coords_2d[1] + object_bbox[1].start
+
+                smoothed_values = smoothed[dilated_coords[:, dilation_axis], dilated_coords[:, -1]]
+
+                refined_coords, probas = thin_coords(coords=dilated_coords, values=smoothed_values)
```
SergeyTsimfer, line 148: Why not name it just `coords`?

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,1056 @@
...
+
+            for idx, object_bbox in enumerate(objects, start=1):
+                # Refined coords: we refine skeletonize effects by applying it on limited area
+                dilation_axis = self.orthogonal_direction
+                dilation_ranges = (max(0, object_bbox[0].start - self.dilation // 2),
+                                   min(object_bbox[0].stop + self.dilation // 2, self.shape[dilation_axis]))
+
+                object_mask = labeled[dilation_ranges[0]:dilation_ranges[1], object_bbox[-1]] == idx
+
+                # Filter out too little components
```
SergeyTsimfer, line 128: Filter out small components
Although I am not even sure that this comment is required: that is a two-line if-statement with a plain ~Engligh language

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,1056 @@
...
+
+                filtered_prototypes.append(prototype)
+
+        return filtered_prototypes
+
+    def prototypes_to_faults(self, field):
+        """ Convert all prototypes to faults. """
+        faults = [Fault(prototype.coords, field=field) for prototype in self.prototypes]
+        return faults
+
```
SergeyTsimfer, line 731: Use double new-lines between different blocks of methods, use triple new-lines between classes

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,1056 @@
...
+class Component:
+    """ Container for extracted connected components.
```
SergeyTsimfer, line 733: That is one component, is not it?

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,1056 @@
...
+        Bounding box.
+    length : int
+        Component length.
+    """
+    def __init__(self, points, slide_idx, bbox=None, length=None):
+        self.points = points
+        self.slide_idx = slide_idx
+
+        self._bbox = bbox
+        self._length = length
```
SergeyTsimfer, line 751: Not used anywhere

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,1056 @@
...
+
+    @property
+    def bbox(self):
+        """ Bounding box. """
+        if self._bbox is None:
+            self._bbox = np.column_stack([np.min(self.coords, axis=0), np.max(self.coords, axis=0)])
+        return self._bbox
+
+    def __len__(self):
+        """ Length. """
```
SergeyTsimfer, line 766: `Number of points in a component`

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,1056 @@
...
+        self._length = length
+
+    @property
+    def coords(self):
+        """ Spatial coordinates in the (ilines, xlines, depths) format."""
+        return self.points[:, :-1]
+
+    @property
+    def bbox(self):
+        """ Bounding box. """
```
SergeyTsimfer, line 760: `3D bounding box`

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,1056 @@
...
+        faults = [Fault(prototype.coords, field=field) for prototype in self.prototypes]
+        return faults
+
+class Component:
+    """ Container for extracted connected components.
+
+    Parameters
+    ----------
+    points : np.ndarray of (N, 4) shape
+        Spatial coordinates and probabilities in the (ilines, xlines, depths, proba) format.
```
SergeyTsimfer, line 738: I think we shoould also add information about the sorting of the arrays: that would help a lot

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,1056 @@
...
+            The most closest component to the splitted.
+        new_components : list of `~.Component` instances
+            Components created from splitted parts.
+        """
+        component = self
+        new_components = []
+
+        # Cut upper part of the component and save it as another item
+        if split_indices[0] is not None:
+            new_component_points = component.points[component.points[:, 2] < split_indices[0]]
```
SergeyTsimfer, line 789: You can create `mask` variable and use it here and use `~mask` to slice `self.points`

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,1056 @@
...
+            Depth values (upper and lower) to split component into parts. If None, the no need in split.
+
+        Returns
+        -------
+        component : `~.Component` instance
+            The most closest component to the splitted.
+        new_components : list of `~.Component` instances
+            Components created from splitted parts.
+        """
+        component = self
```
SergeyTsimfer, line 784: I would just use `self` in both conditions

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,1056 @@
...
+            new_component_bbox = component.bbox.copy()
+            new_component_bbox[2, 1] = max(0, split_indices[0] - 1)
+
+            new_component = Component(points=new_component_points, slide_idx=self.slide_idx, bbox=new_component_bbox)
+
+            new_components.append(new_component)
+
+            # Extract suitable part
+            component.points = component.points[component.points[:, 2] >= split_indices[0]]
+            component.bbox[2, 0] = split_indices[0]
```
SergeyTsimfer, line 800: I would first modify `self` and then add new components to the list: that is better aligned with logic (and documentation, for that matter)

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,1056 @@
...
+        ||||||
+
+         - where | means one prototype points, and . - other prototype points.
+
+        Parameters
+        ----------
+        border_threshold : int
+            Minimal amount of points out of borders overlap to decide that prototypes are not close.
+        """
+        # Presort objects by other valuable axis for early stopping
```
SergeyTsimfer, line 532: What does `valuable axis` mean?

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,1056 @@
...
+        for i, prototype_1_idx in enumerate(prototypes_order):
+            prototype_1 = self.prototypes[prototype_1_idx]
+
+            for prototype_2_idx in prototypes_order[i+1:]:
+                prototype_2 = self.prototypes[prototype_2_idx]
+                # Check that prototypes are embedded
+                if (prototype_1.bbox[sort_axis, 1] < prototype_2.bbox[sort_axis, 0]):
+                    break
+
+                is_embedded, is_second_inside_first = bboxes_embedded(prototype_1.bbox, prototype_2.bbox, margin=margin)
```
SergeyTsimfer, line 548: I would name the second flag something along the lines of `swap`:
```python
is_inside, swap = ...
coords = p1.coords if swap is False else p2.coords
other = p1 if swap is False else p2
```

seismiqb/labels/fault/coords_utils.py
```diff
@@ -0,0 +1,271 @@
+""" Helpers for the natural 3-dimensional space coordinates processing. """
```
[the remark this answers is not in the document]
HollowPrincess: Natural numbers

seismiqb/labels/fault/extractor.py
HollowPrincess: To convert to integer type

seismiqb/labels/fault/coords_utils.py
HollowPrincess: **zeros**: 1min 40s  /  5min 13s
**empty**: 1min 42s  / 5min 20s

seismiqb/labels/fault/extractor.py
HollowPrincess: I want to add it later (in the separate PR) for chunking option.

seismiqb/labels/fault/coords_utils.py
```diff
@@ -0,0 +1,271 @@
...
+        For results accuracy, provide step = 1.
+    axis : {0, 1}
+        Axis along which to find distances.
+    max_threshold : int, float or None
+        Early stopping: threshold for max distance value.
+        If the max value is more than this value, then no sense in finding more accurate results
+        because coordinates are too far from each other.
+        If None, then o threshold is applied.
+    """
+    min_distance = max_threshold or 100
```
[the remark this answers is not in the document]
HollowPrincess: I made `max_threshold = 1000` and `min_distance = max_threshold` - timings are same

seismiqb/labels/fault/coords_utils.py
HollowPrincess: I checked - no difference

seismiqb/labels/fault/coords_utils.py
```diff
@@ -0,0 +1,271 @@
...
+        dilated_coords[start_idx:end_idx, axis] += i - dilate//2
+
+    # Clip to the valid values
+    if max_value is not None:
+        dilated_coords = dilated_coords[(dilated_coords[:, axis] >= 0) & (dilated_coords[:, axis] <= max_value)]
+    else:
+        dilated_coords = dilated_coords[dilated_coords[:, axis] >= 0]
+
+    # Get sorted unique values
+    dilated_coords = np.unique(dilated_coords, axis=0)
```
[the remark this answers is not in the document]
HollowPrincess: Hundreds, maybe thousands

Now I have no need in dorted values, so:
```
dilated_coords = pd.DataFrame(dilated_coords).drop_duplicates().values
```
Takes 2min 27s instead of 1min 40s.

seismiqb/labels/fault/coords_utils.py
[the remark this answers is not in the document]
HollowPrincess: `np.in1d` doesn't noticeably improve performance: 1min 43s instead of 1min 40s

seismiqb/labels/fault/coords_utils.py
```diff
@@ -0,0 +1,257 @@
...
+    Under the hood, this function is a groupby max along the depth axis.
+
+    Parameters
+    ----------
+    coords : np.ndarray of (N, 3) shape
+        Coordinates for thinning. Sorting is not required.
+    values : np.ndarray of (N, 1) shape
+        Values corresponding to coordinates to decide which one should last for each depth (last column in coords).
+    """
+    order = np.argsort(coords[:, -1])[::-1]
```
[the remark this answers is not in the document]
HollowPrincess: I think you are right and there is no need in reverse sorting, I will check it.

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,1245 @@
...
+            - filter all unsuitable prototypes.
+
+        Parameters
+        ----------
+        prolongate_in_depth : bool
+            Whether to maximally prolongate faults in depth or not.
+            If True, then surfaces will be tall and thin.
+            If False, then surfaces will be more longer for `self.direction` than for depth axis.
+        concat_iters : int
+            Maximal amount of connected component concatenation operations which are include concat along
```
AlexeyKozhevin, line 751:
```suggestion
            Maximal amount of connected prototypes concatenation operations which are include concat along
```

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,1357 @@
...
+                continue
+
+            # Check bboxes intersection
+            if not bboxes_intersected(dilated_bbox, other_component.bbox, axes=(self.orthogonal_direction, 2)):
+                continue
+
+            # Check closeness of some points (as depth-wise distances)
+            # Faster then component overlap, but not so accurate
+            overlap_depths = (max(component.bbox[2, 0], other_component.bbox[2, 0]),
+                                    min(component.bbox[2, 1], other_component.bbox[2, 1]))
```
SergeyTsimfer, line 352: Indentation

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,1357 @@
...
+
+        # Presort objects by overlap axis for early stopping
+        sort_axis = overlap_axis
+        prototypes_starts = np.array([prototype.bbox[sort_axis, 0] for prototype in self.prototypes])
+        prototypes_order = np.argsort(prototypes_starts)
+        reodered_prototypes = [self.prototypes[idx] for idx in prototypes_order]
+
+        new_prototypes = []
+
+        for i, prototype_1 in enumerate(reodered_prototypes):
```
SergeyTsimfer, line 472: When you perform `_find_closest_component`, you begin by getting some candidates that can be merged, and the select the best one out of them.

Maybe, we should employ similar approach in this method? Get list of `prototype_2`, that can be merged to the current `prototype_1`, and then select "the best" (in some sense) to actually merge?

SergeyTsimfer (approved):

seismiqb/labels/fault/extractor.py
```diff
@@ -0,0 +1,1357 @@
...
+
+        return closest_component, prototype_split_indices
+
+    def _add_new_components(self, components):
+        """ Add new components into the container.
+
+        New items are created after splitting.
+        """
+        for component in components:
+            if len(component) > self.component_len_threshold:
```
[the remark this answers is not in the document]
HollowPrincess: This attribute is for speed-up, we don't use it.
But if we want make extraction faster, then components with ~3 points are not necessary.

PR #254, opened 2023-02-28
Small fixes; close issues

SergeyTsimfer:


HollowPrincess:
Please, add `dump_charisma` fix in not-'points' case (`np.isnan` instead of `== np.nan`).

PR #257, opened 2023-03-31
Add sections loading

AlexeyKozhevin:
image
image
image


seismiqb/geometry/segy.py
```diff
@@ -569,3 +606,90 @@ def _collect_stats_chunk(data,
...
+    Returns
+    -------
+    locations : numpy.ndarray
+        array of shape (N, 4) and dtype float32 with coordinates for section traces.
+    """
+    direction = loc_b - loc_a
+    distance = np.power(direction, 2).sum() ** 0.5
+    locations = np.empty((int(np.ceil(distance)) + 1, 2), dtype=np.float32)
+    for i in [0, 1]:
+        locations[:, i] = np.linspace(loc_a[i], loc_b[i], int(np.ceil(distance)) + 1)
```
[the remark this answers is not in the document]
AlexeyKozhevin: That's nice to be fast. `np.linspace` under `njit` doesn't work with tuples
image

seismiqb/geometry/segy.py
```diff
@@ -569,3 +606,90 @@ def _collect_stats_chunk(data,
...
+    ceil, floor = np.ceil(locations), np.floor(locations)
+    support = np.empty((len(locations), 4, 2), dtype='int32')
+    support[:, 0] = floor
+    support[:, 1, 0] = floor[:, 0]
+    support[:, 1, 1] = ceil[:, 1]
+    support[:, 2, 0] = ceil[:, 0]
+    support[:, 2, 1] = floor[:, 1]
+    support[:, 3] = ceil
+
+    weights = np.sum((np.expand_dims(locations, 1) - support) ** 2, axis=-1) ** 0.5
```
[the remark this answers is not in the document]
AlexeyKozhevin: ` keepdims=True` doesn't work under njit but we need it:
image

seismiqb/geometry/segy.py
```diff
@@ -569,3 +612,90 @@ def _collect_stats_chunk(data,
...
+    weights = 1 / weights
+    for i in prange(len(weights)):
```
EvgeniyS99, line 668: I think you could use:
```python
weights_sum = np.expand_dims(weights.sum(axis=-1), 1)
weights = weights / weights_sum
```
And it will work under the njit decorator

seismiqb/geometry/segy.py
```diff
@@ -569,3 +612,90 @@ def _collect_stats_chunk(data,
...
+        weights[i] = weights[i] / sum(weights[i])
+    return support, weights
+
```
EvgeniyS99, line 671: I think you could use:
```python
weights_sum = np.expand_dims(weights.sum(axis=-1), 1)
weights = weights / weights_sum
```
And it will work under the njit decorator

seismiqb/geometry/base.py
```diff
@@ -778,6 +777,120 @@ def show_slide(self, index, axis=0, zoom=None, plotter=plot, **kwargs):
...
+        labels = np.array(list(map('\n'.join, indices.astype('int32').astype(str))))[xticks % section.shape[0]]
+
+        plt[0].ax.set_xticks(xticks[:-1])
+        plt[0].ax.set_xticklabels(labels[:-1])
+
+        if linecolor:
+            for pos in nodes:
+                plt[0].ax.plot([pos, pos], [0, section.shape[1]], color=linecolor, linewidth=linewidth)
+
+        plt.redraw()
```
SergeyTsimfer, line 848: `return plotter` so that we can add to it after the method call
Also, is the `redraw` call needed?
Also, current way would not allow to use `savepath` correctly: figure will be saved at the first call to `plotter`, and further additions (vertical lines in your case) would not be saved

seismiqb/geometry/segy.py
```diff
@@ -525,6 +526,48 @@ def get_optimal_axis(self, locations=None, shape=None):
...
+                2D array with loaded and interpolated traces of section.
+            indices : numpy.ndarray
+                Float coordinates of section traces.
+            nodes : numpy.ndarray
+                Positions of node traces (from `locations`) in `traces` array.
+        """
+        dtype = dtype or self.dtype
+
+        indices = []
+        for i in range(1, len(locations)):
```
SergeyTsimfer, line 553: I would prefer `for loc_a, loc_b in zip(locations[:-1], locations[1:])`

seismiqb/geometry/segy.py
```diff
@@ -569,3 +612,90 @@ def _collect_stats_chunk(data,
 
     return (min_vector, max_vector, mean_vector, var_vector,
             min_matrix, max_matrix, mean_matrix, var_matrix)
+
+@njit
+def get_line_traces(loc_a, loc_b):
```
SergeyTsimfer, line 617: Maybe, call it something like: `get_line_rasterization(start, stop)` or `get_line_coordinates(start, stop`? get traces sounds like it loads actual trace data, which is not the case

seismiqb/geometry/base.py
AlexeyKozhevin: Yes,  we need `redraw` to add lines on map.

PR #258, opened 2023-04-04
Spatial check points

EvgeniyS99:
The **spatial_check_points** function was fixed. The previous version of this function didn't support the **threshold** parameter, because  when counting the present points the dead traces were considered. 

The **lines_to_ordinals** function was fixed. The in-place division was replaced with the true division.


seismiqb/samplers.py
```diff
@@ -668,7 +668,7 @@ def spatial_check_points(points, matrix, crop_shape, i_mask, x_mask, threshold):
             present_points, running_sum = np.int32(0), np.int32(0)
 
             for value in np.nditer(sliced):
-                if value >= 0:
+                if value > 0: # the previous condition was `value >= 0` and the dead traces were considered
```
SergeyTsimfer, line 671: No reason to mention previous code

seismiqb/samplers.py
```diff
@@ -44,7 +44,7 @@ def _make_locations(self, field, points, matrix, crop_shape, ranges, threshold,
 
         # Keep only points, that can be a starting point for a crop of given shape
         i_mask = ((ranges[:2, 0] <= points[:, :2]).all(axis=1) &
-                  ((points[:, :2] +   crop_shape[:2]) <= ranges[:2, 1]).all(axis=1))
+                  ((points[:, :2] + crop_shape[:2]) <= ranges[:2, 1]).all(axis=1))
```
SergeyTsimfer, line 47: Why? There is a reason for those whitespaces

seismiqb/geometry/base.py
```diff
@@ -407,7 +407,7 @@ def lines_to_ordinals(self, array):
         # Depth to units
         if array.shape[1] == self.index_length + 1:
             array[:, self.index_length] -= self.delay
-            array[:, self.index_length] /= self.sample_interval
+            array[:, self.index_length] = array[:, self.index_length] / self.sample_interval
```
SergeyTsimfer, line 410: Make sure that the dtype is the same as before this change

seismiqb/samplers.py
EvgeniyS99: Sorry, but I thought there was an extra space in this line, cause it's missing in the next line. 
image


> Why? There is a reason for those whitespaces

seismiqb/geometry/base.py
EvgeniyS99: > Make sure that the dtype is the same as before this change

The returned values of this function are passed to the following function:
image,
Which implies that the returned values must be integers. Does it mean that I should apply integer division?

SergeyTsimfer (approved):

PR #261, opened 2023-04-10
Update accumulator

SergeyTsimfer:
This PR:
- speeds up some of the inner computations of `RegularGrid` and makes `strides` parameter parse to `crop_shape` by default
- adds SHM type for all accumulators: memory for data is shared across process and can be accessed by its name

### DIFF
--- a/seismiqb/geometry/base.py
+++ b/seismiqb/geometry/base.py
@@ -5,7 +5,10 @@ from textwrap import dedent
 from contextlib import contextmanager
 
 import numpy as np
+
+import cv2
 from scipy.interpolate import interp1d
+from scipy.ndimage import binary_erosion
 
 from .benchmark_mixin import BenchmarkMixin
 from .conversion_mixin import ConversionMixin
@@ -196,9 +199,8 @@ class Geometry(BenchmarkMixin, CacheMixin, ConversionMixin, ExportMixin, MetricM
     def __getstate__(self):
         self.reset_cache()
         state = self.__dict__.copy()
-        for name in ['loader', 'axis_to_projection']:
-            if name in state:
-                state.pop(name)
+        for name in ['loader', 'axis_to_projection'] + self.PRESERVED_LAZY_CACHED:
+            state.pop(name, None)
         return state
 
     def __setstate__(self, state):
@@ -597,7 +599,7 @@ class Geometry(BenchmarkMixin, CacheMixin, ConversionMixin, ExportMixin, MetricM
     @property
     def ngbytes(self):
         """ Size of instance in gigabytes. """
-        return self.nbytes / (1024**3)
+        return self.nbytes / (1024 ** 3)
 
 
     # Attribute retrieval. Used by `Field` instances
@@ -989,3 +991,40 @@ class Geometry(BenchmarkMixin, CacheMixin, ConversionMixin, ExportMixin, MetricM
     def locations_to_shape(locations):
         """ Compute shape of a location. """
         return tuple(slc.stop - slc.start for slc in locations)
+
+    def get_slide_mask(self, index, axis=0, kernel_size=9, threshold=None, erosion=11):
+        """ Get mask with dead pixels on a given slide.
+        Under the hood, we compute ptp value for each pixel, and deem everything lower than `threshold` be a dead pixel.
+
+        Parameters
+        ----------
+        index : int, str
+            Index of the slide to show.
+            If int, then interpreted as the ordinal along the specified axis.
+            If `'random'`, then we generate random index along the axis.
+            If string of the `'#XXX'` format, then we interpret it as the exact indexing header value.
+        axis : int
+            Axis of the slide.
+        kernel_size : int
+            Window size for computations.
+        threshold : number
+            Minimum ptp value to consider a pixel to be a dead one.
+        erosion : int
+            Amount of binary erosion for postprocessing.
+
+        Returns
+        -------
+        mask : np.ndarray
+            Boolean mask with 1`s at dead pixels and 0`s at alive ones.
+        """
+        threshold = threshold or self.std / 10
+        slide = self.load_slide(index=index, axis=axis,)
+
+        kernel = np.ones((kernel_size, kernel_size), dtype=slide.dtype)
+        ptps = cv2.dilate(slide, kernel) - cv2.erode(slide, kernel)
+        mask = ptps <= threshold
+
+        if erosion:
+            kernel = np.ones((erosion, erosion), dtype=mask.dtype)
+            mask = binary_erosion(mask, structure=kernel, border_value=True)
+        return mask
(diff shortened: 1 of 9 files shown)


seismiqb/utils/accumulator3d.py
```diff
@@ -52,8 +55,8 @@ class Accumulator3D:
     kwargs : dict
         Other parameters are passed to HDF5 dataset creation.
     """
-    def __init__(self, shape=None, origin=None, orientation=0, dtype=np.float32, transform=None, path=None,
-                 dataset_kwargs=None, **kwargs):
+    def __init__(self, shape=None, origin=None, orientation=0, dtype=np.float32, transform=None,
+                 type_=None, path=None, dataset_kwargs=None, **kwargs):
```
AlexeyKozhevin, line 59: Rename `type_` to `fmt`

seismiqb/utils/functions.py
```diff
@@ -20,6 +22,10 @@ def select_printer(printer):
     return print
 
 
+def generate_string(size=10, chars=string.ascii_uppercase + string.digits):
+    """ Generate random string of given size. """
+    return ''.join(random.choice(chars) for _ in range(size))
```
HollowPrincess, line 27: Can we have name collisions because of this? Or it is not cruicial?

Maybe we need a numbers generator as a prefix?

seismiqb/geometry/base.py
```diff
@@ -861,3 +863,40 @@ def process_limits(self, limits):
...
+        erosion : int
+            Amount of binary erosion for postprocessing.
+
+        Returns
+        -------
+        mask : np.ndarray
+            Boolean mask with 1`s at dead pixels and 0`s at alive ones.
+        """
+        threshold = threshold or self.std / 10
+        slide = self.load_slide(index=index, axis=axis,)
```
AlexeyKozhevin, line 893:
```suggestion
        slide = self.load_slide(index=index, axis=axis)
```

seismiqb/geometry/base.py
```diff
@@ -861,3 +863,40 @@ def process_limits(self, limits):
...
+
+        kernel = np.ones((kernel_size, kernel_size), dtype=slide.dtype)
+        ptps = cv2.dilate(slide, kernel) - cv2.erode(slide, kernel)
```
AlexeyKozhevin, line 896: Doesn't work with `int8` slides:
Screenshot 2023-04-13 at 18 48 03

AlexeyKozhevin (changes requested):

seismiqb/utils/accumulator3d.py
```diff
@@ -85,24 +103,97 @@ def reorder(self, sequence):
...
+            placeholder = np.ndarray(buffer=shm.buf, shape=self.shape, dtype=dtype)
+            placeholder[:] = fill_value
+
+            self.shm_data[name] = [shm, dtype]
+
+        self.placeholders.append(name)
         setattr(self, name, placeholder)
 
-    def remove_placeholder(self, name=None):
+    def remove_placeholder(self, name=None, unlink=False):
```
HollowPrincess, line 137: Looks like `unlink` is a special 'shm' kwarg and doesn't coresspond to other types.
It is unimportant, but I propose to rename `unlink` to `kwargs` and pop it in 'shm' case.

seismiqb/utils/accumulator3d.py
```diff
@@ -85,24 +103,97 @@ def reorder(self, sequence):
...
+        if self.type == 'shm':
+            for name in self.placeholders:
+                self.remove_placeholder(name, unlink=unlink)
+
+    def __getstate__(self):
+        """ Store state of an instance. Remove file handlers and shared memory objects. """
+        state = copy(self.__dict__)
+        if self.type in ['hdf5', 'qhdf5', 'zarr']:
+            for name in self.placeholders:
+                state[name] = None
```
HollowPrincess, line 165: Why we doesn't do the same for 'numpy' placeholder?

seismiqb/utils/accumulator3d.py
```diff
@@ -224,25 +315,36 @@ def int8_to_prediction(array):
...
         array /= 255
         return array
 
+    @staticmethod
+    def prediction_to_uint16(array):
+        """ Convert a float array with values in [0.0, 1.0] to an uint16 array with values in [0, 255].
+        Useful for accumulators that need to keep track of sum of values.
+        """
+        array *= 255
+        return array.astype(np.uint16)
```
HollowPrincess, line 335: What about `convert_prediction(array, dtype)` method?

seismiqb/utils/accumulator3d.py
```diff
@@ -305,15 +422,57 @@ def _aggregate(self):
...
         # Cleanup
-        self.remove_placeholder('counts')
+        self.remove_placeholder('counts', unlink=True)
+
+
+class StdAccumulator3D(Accumulator3D):
+    """ Accumulator that takes std value of overlapping crops. """
+    def __init__(self, shape=None, origin=None, dtype=np.float32, transform=None, path=None, **kwargs):
+        if dtype not in [np.float32, np.float64]:
+            raise ValueError('Dtype should be floating for STD accumulator!')
```
HollowPrincess, line 440: I think there is should be a check for float subtype on the 439 line or the error message should contain supported dtypes.

seismiqb/utils/accumulator3d.py
```diff
@@ -305,15 +422,57 @@ def _aggregate(self):
...
+                self.data[i] **= 1/2
+
+        elif self.type in ['numpy', 'shm']:
+            self.counts[self.counts == 0] = 1
+            self.data /= self.counts
+            self.sum /= self.counts
+
+            self.data -= self.sum ** 2
+            self.data **= 1/2
+
```
HollowPrincess, line 472: Maybe it would be better to reduce these lines of code with one function `_std_aggregator(counts, data, sum)`?

AlexeyKozhevin (approved):

seismiqb/grids.py
```diff
@@ -382,14 +394,15 @@ def __init__(self, grid, size, overlap):
 
         size_i, size_x = size
         overlap_i, overlap_x = overlap
+        float_types = (float, np.float16, np.float32, np.float64, np.float128)
```
HollowPrincess, line 397:
```python
np.issubdtype(type(variable), np.floating)
```

HollowPrincess (approved):

seismiqb/geometry/base.py
```diff
@@ -989,3 +991,41 @@ def process_limits(self, limits):
     def locations_to_shape(locations):
         """ Compute shape of a location. """
         return tuple(slc.stop - slc.start for slc in locations)
+
+    def get_slide_mask(self, index, axis=0, kernel_size=9, threshold=None, erosion=11):
```
AlexeyKozhevin, line 995: The function works only with the whole slide but sometimes its needed to proccess some crop from it. Add `locations/ranges/smth`

PR #266, opened 2023-04-26
Fix weighted accumulator

EvgeniyS99:


seismiqb/utils/accumulator3d.py
```diff
@@ -401,6 +401,9 @@ def _update(self, crop, location):
         # Weights matrix for the incoming crop
         if self.crop_weights is None:
             self.crop_weights = self.weights_function(crop)
+        
+        if self.crop_weights.shape != crop.shape:
```
SergeyTsimfer, line 405: Why would you ever have crops that are not the same shape?

PR #269, opened 2023-05-16
Add proba-int conversion

AlexeyKozhevin:


seismiqb/utils/conversion.py
```diff
@@ -34,3 +35,15 @@ def repack_hdf5(path, dst_path=None, projections = (0, ), transform=None, dtype=
     if inplace:
         os.remove(path)
         os.rename(dst_path, path)
+
+def proba_to_int(x, dtype='int8'):
```
SergeyTsimfer, line 39: I would rather make a few template-like functions with hardcoded dtype: `proba_to_int8`, `proba_to_int16` etc. Reason being the ability to pass those functions easily as transforms to `Accumulator3d` and `export_array` functions: `accumulator = Accumulator3d(..., transform=dtype_conversion.proba_to_int8)` instead of using `partial`

seismiqb/utils/conversion.py
```diff
@@ -1,6 +1,7 @@
 """ Utils for conversion. """
 
 import os
+import numpy as np
```
SergeyTsimfer, line 4: I would rather make it a separate file: conversion between formats and numeric types is sufficiently different

seismiqb/utils/conversion.py
SergeyTsimfer: Those generic functions can be kept as well

SergeyTsimfer (approved):

PR #274, opened 2023-06-22
Small updates

HollowPrincess:
* Sample (depth) to time conversion
* Add origin into horizon extractor

### DIFF
--- a/seismiqb/geometry/utils.py
+++ b/seismiqb/geometry/utils.py
@@ -1,0 +1,10 @@
+""" Useful functions, related to geometry. """
+
+
+def time_to_depth(time, geometry):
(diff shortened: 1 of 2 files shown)


seismiqb/geometry/utils.py
```diff
@@ -0,0 +1,10 @@
+""" Useful functions, related to geometry. """
+
+
+def time_to_depth(time, geometry):
+    """ Convert time (in ms) into geometry depth value. """
```
SergeyTsimfer, line 5: You are not converting to `depth`; you are converting to `samples`, which are measuring time as well
That is a moment where I regret naming the z-axis `depth` and not `samples`

SergeyTsimfer (approved):

PR #277, opened 2023-07-05
Handler for read/dump matrices in IRAP ASCII data format

HollowPrincess:


seismiqb/geometry/irap_handler.py
```diff
@@ -0,0 +1,151 @@
+""" IRAP ASCII files handler. """
```
SergeyTsimfer, line 1: I would rather move this file to `utils`

seismiqb/geometry/irap_handler.py
```diff
@@ -0,0 +1,151 @@
...
+import numpy as np
+
+
+
+class _IRAPHandler:
+    """ IRAP ASCII files handler for reading and dumping matrices. """
+    # File format constants
+    fill_value = 9999900.0
```
SergeyTsimfer, line 9: constans should be in CAPITAL LETTERS

seismiqb/geometry/irap_handler.py
```diff
@@ -0,0 +1,151 @@
...
+        path : str
+            Path to IRAP ASCII data file.
+        geometry : instance of :class:`~.Geometry`
+            Geometry that the file corresponds to.
+        """
+        # Extract data from file
+        with open(path, mode='r') as irap_file:
+            data = irap_file.read()
+
+        values = data.split() # We can't use `np.loadtxt` because some lines can be less than 6 elements
```
SergeyTsimfer: Pandas can work with csv-like files with missing values: that would be a lot faster

seismiqb/geometry/irap_handler.py
```diff
@@ -0,0 +1,151 @@
...
+            ranges = geometry.cdp_to_lines(ranges)
+            ranges = geometry.lines_to_ordinals(ranges)
+
+            ranges = np.ceil(ranges).astype(np.int32)
+
+            start = (ranges[0, 0], ranges[0, 1])
+            stop = (ranges[1, 0] + 1, ranges[1, 1] + 1)
+
+            # Find step
+            rotation_matrix = geometry.rotation_matrix.round(1).astype(np.int32)
```
SergeyTsimfer, line 52: While values of `CDP_*` are `int`, their increment can be float: for example, values of `100, 113, 125, 138, 150` would mean that the actual step is 12.5.

I would try to avoid explicit rounds / usage of values if possible

seismiqb/geometry/irap_handler.py
```diff
@@ -0,0 +1,151 @@
...
+            if rotation_matrix[0, 0] != 0:
+                dx, dy = rotation_matrix[0, 0], rotation_matrix[1, 1] # Need to be properly checked
+            else:
+                dx, dy = rotation_matrix[1, 0], rotation_matrix[0, 1]
+
+            step = (round(float(header[3])/dx), round(float(header[2])/dy))
+
+            # Create matrix
+            matrix = np.empty(geometry.shape[:2], dtype=np.float32)
+            matrix[:, :] = np.nan
```
SergeyTsimfer, line 63: looks like `np.full(shape, fill_value=np.nan, dtype=np.float32)`

seismiqb/geometry/irap_handler.py
```diff
@@ -0,0 +1,151 @@
...
+            mgrid = np.mgrid[start[0]:stop[0]:step[0], start[1]:stop[1]:step[1]]
+            matrix[mgrid[0], mgrid[1]] = np.array(values, dtype=np.float32).reshape(ncols, nrows)
```
SergeyTsimfer, line 65: why can't you directly use start/stop/step slice here?

seismiqb/geometry/irap_handler.py
```diff
@@ -0,0 +1,151 @@
...
+            # XMIN    XMAX   YMIN  YMAX
+            f"{geometry.headers['CDP_X'].min():.6f} {geometry.headers['CDP_X'].max():.6f} "
+            f"{geometry.headers['CDP_Y'].min():.6f} {geometry.headers['CDP_Y'].max():.6f}\n"
+            # NX   ROTATION(CCW)    ROTX   ROTY
+            f"{matrix.shape[1]} {0:.6f} {geometry.headers['CDP_X'].min():.6f} {geometry.headers['CDP_Y'].min():.6f}\n"
+            "0   0   0   0   0   0   0"
+        )
+        return header
+
+IRAPHandler = _IRAPHandler
```
SergeyTsimfer: Why would you need that?

seismiqb/geometry/irap_handler.py
HollowPrincess: You are right:

**baseline**: 987 ms ± 2.61 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
**pandas** : 337 ms ± 1.62 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

seismiqb/utils/irap_handler.py
```diff
@@ -0,0 +1,215 @@
...
+        format : {'cdp', 'ordinal'}
+            Coordinate system in which to evaluate step.
+        header : str or None
+            Header from which to take information about cdp coordinates step.
+            Used only for conversion from 'cdp' to 'ordinal' format.
+        """
+        # Eval step in cdp coordinates format
+        rotation_matrix = geometry.rotation_matrix
+
+        if rotation_matrix[0, 0] > rotation_matrix[0, 1]: # Choose main diagonal
```
AlexeyKozhevin, line 207: Why not a `rotation_matrix.sum(axis=1)`?
As I understand, `step` is as difference in world coordinate between two pairs of points (in i-x coordinates): 
1) (0, 0) and (1, 0)
2) (0, 0) and (0, 1).

In that case, I guess, it should be `(rotation_matrix ** 2).sum(axis=1) ** 0.5`

PR #278, opened 2023-07-06
Replace `np.take` with basic indexing in `get_crop_mask`

EvgeniyS99:
`np.take` works too long on subcubes of the original cube. 
For example:

image

### DIFF
--- a/seismiqb/geometry/base.py
+++ b/seismiqb/geometry/base.py
@@ -15,7 +15,7 @@ from .conversion_mixin import ConversionMixin
 from .export_mixin import ExportMixin
 from .metric_mixin import MetricMixin
 
-from ..utils import SQBStorage, lru_cache, CacheMixin, TransformsMixin, select_printer, transformable
+from ..utils import SQBStorage, lru_cache, CacheMixin, TransformsMixin, select_printer, transformable, take_along_axis
 from ..plotters import plot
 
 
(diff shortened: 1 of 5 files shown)


seismiqb/geometry/base.py
```diff
@@ -1059,7 +1059,12 @@ def get_crop_mask(self, locations, axis=0, kernel_size=9, threshold=None, erosio
         dilation_kernel = np.ones((dilation, dilation), dtype=array.dtype)
 
         for i in range(array.shape[axis]):
-            slide = array.take(i, axis=axis)
+            if axis == 0:
+                slide = array[i, ...]
```
AlexeyKozhevin, line 1063:
```suggestion
                slide = array[i]
```

seismiqb/geometry/base.py
```diff
@@ -1059,7 +1059,12 @@ def get_crop_mask(self, locations, axis=0, kernel_size=9, threshold=None, erosio
...
+            elif axis == 1:
+                slide = array[:, i, :]
```
AlexeyKozhevin, line 1065:
```suggestion
                slide = array[:, i]
```

seismiqb/geometry/base.py
SergeyTsimfer: `array[i, :]` and `array[:, i]` so that they have the same form

PR #283, opened 2023-08-04
Speed up `get_crop_mask`

AlexeyKozhevin:


seismiqb/geometry/base.py
```diff
@@ -1151,16 +1156,28 @@ def get_crop_mask(self, locations, axis=0, kernel_size=9, threshold=None, erosio
...
-            mask_ = ptps <= threshold
-            if erosion:
-                mask_ = binary_erosion(mask_, structure=erosion_kernel, border_value=True)
-            if dilation:
-                mask_ = cv2.dilate(mask_.astype('uint8'), dilation_kernel).astype('bool')
-            slc = [slice(None)] * 3
-            slc[axis] = i
-            mask[tuple(slc)] = mask_
+        futures = []
+        with ThreadPoolExecutor(max_workers=max_workers or array.shape[axis]) as executor:
```
SergeyTsimfer, line 1160: Move callback creation here, so that the usage of Executor is similar to its usage in other places in our code: for example, `geometry.segy.collect_stats`

seismiqb/geometry/base.py
```diff
@@ -1151,16 +1156,28 @@ def get_crop_mask(self, locations, axis=0, kernel_size=9, threshold=None, erosio
...
+                slide = take_along_axis(array, i, axis)
+
+                future = executor.submit(_get_mask, slide, ptp_kernel, threshold, erosion_kernel, dilation_kernel, erosion, dilation)
+                future.add_done_callback(partial(_callback, mask=mask, i=i, axis=axis))
 
+                futures.append(future)
+            wait(futures)
         return mask
+
+def _get_mask(array, ptp_kernel, threshold, erosion_kernel, dilation_kernel, erosion=11, dilation=60):
```
SergeyTsimfer, line 1171: Can be a part of the class, as it is used in `ThreadPoolExecutor`

PR #284, opened 2023-08-09
Update attributes for evaluation in points

HollowPrincess:


seismiqb/labels/horizon/attributes.py
```diff
@@ -382,6 +400,39 @@ def get_cube_values(self, window=1, offset=0, chunk_size=256, src_geometry=None,
...
+                            slice(depth - low, min(depth + high, self.field.depth)))
+                location, _ = geometry.process_key(location)
+                data_chunk = geometry.load_crop(location, use_cache=False)
+
+                result[idx, :] = data_chunk[0, 0, :]
+
+        if apply_float_correction:
+            result = self._apply_float_correction(result, coords)
+
+        return result
```
SergeyTsimfer, line 427: Make the output 3D here: either (N, 1, D) or (1, N, D). That would make unnecessary all of the code changes in other methods

seismiqb/labels/horizon/attributes.py
```diff
@@ -103,6 +104,13 @@ def matrix_normalize(self, matrix, mode):
         elif mode == 'mean-std':
             mean, std = np.nanmean(values), np.nanstd(values)
             matrix = (matrix - mean) / std
+        elif mode == 'q':
```
SergeyTsimfer, line 107: Now as I think about it
Why would not we use `Normalizer` class, introduced by @AlexeyKozhevin, here?

seismiqb/labels/horizon/attributes.py
```diff
@@ -594,13 +654,16 @@ def get_property(self, src, **_):
             raise ValueError(f'Unknown `src` {src}. Expected a matrix-property or one of {aliases}.')
         return data
 
-    @lru_cache(maxsize=1, apply_by_default=False, copy_on_return=True)
+    @lru_cache(maxsize=1, apply_by_default=False, copy_on_return=True,
+               disable_cache=[lambda kwargs: kwargs.get('coords', None) is not None])
     @transformable
-    def get_instantaneous_amplitudes(self, window=11, offset=0, **kwargs):
+    def get_instantaneous_amplitudes(self, coords=None, window=11, offset=0, **kwargs):
```
SergeyTsimfer, line 660: I would move `coords` parameter to the end of the signature: important parameters for attribute computation are `window, ...`, and `coords` is more of a implementation detail

seismiqb/labels/horizon/attributes.py
```diff
@@ -650,69 +716,77 @@ def get_metric(self, metric='support_corrs', supports=50, agg='nanmean', **kwarg
...
         spectral =  compute_spectral_decomposition(amplitudes, frequencies=frequencies, wavelet=wavelet,
                                                    sample_rate=sample_rate, method=method)
-        spectral = spectral[..., window // 2].transpose(1, 2, 0)
 
-        if normalize:
-            q = np.nanquantile(spectral, (0.01, 0.99), axis=(0, 1)).reshape(2, 1, 1, -1)
-            spectral = np.clip(spectral, a_min=q[0], a_max=q[1])
-            spectral -= q[0]
-            spectral /= (q[1] - q[0])
+        spectral = spectral[...,  window // 2].transpose(1, 2, 0)
```
SergeyTsimfer, line 789: If the idea is to apply `normalization` via decorator, then it would give different results (because of the different order)

seismiqb/utils/cache.py
```diff
@@ -272,7 +276,17 @@ def wrapper(*args, **kwargs):
             # if a bound method, get class instance from function else from arguments
             instance = func.__self__ if self.is_method else args[0]
 
+            # Parse use_cache
             use_cache = kwargs.pop('use_cache', self.apply_by_default)
+            disable_cache = kwargs.pop('disable_cache', self.disable_cache)
+
+            if use_cache:
+                for disable in disable_cache:
```
SergeyTsimfer, line 284: Make it one callable, not a list of them
If user wanted to make multiple conditions, they could be fit into that one callable

